PR #4, opened 2018-05-29
Replace Parse functions with JSON marshalling

GanShun:
This allows us to prepare to serialize out metadata into files for easy parsing, and we can remove the Summary calls as I've made the Json print out similar fields to the Summary calls

### DIFF
--- a/uefi/firmwarevolume.go
+++ b/uefi/firmwarevolume.go
@@ -37,16 +37,16 @@ type Block struct {
 // FirmwareVolumeFixedHeader contains the fixed fields of a firmware volume
 // header
 type FirmwareVolumeFixedHeader struct {
-	Zeros          [16]uint8
+	_              [16]uint8
 	FileSystemGUID [16]uint8
 	Length         uint64
 	Signature      uint32
 	AttrMask       uint8
 	HeaderLen      uint16
 	Checksum       uint16
-	Reserved       [3]uint8
+	Reserved       [3]uint8 `json:"-"`
 	Revision       uint8
-	Unused         [3]uint8
+	_              [3]uint8
 }
 
 // FirmwareVolume represents a firmware volume. It combines the fixed header and
@@ -56,24 +56,14 @@ type FirmwareVolume struct {
 	// there must be at least one that is zeroed and indicates the end of the
 	// block list
 	Blocks []Block
+
+	// Variables not in the binary for us to keep track of stuff/print
+	guidString string
+	guidName   string
 }
 
 // Summary prints a multi-line representation of a FirmwareVolume object
 func (fv FirmwareVolume) Summary() string {
-	var (
-		guidName, guidString string
-		ok                   bool
-	)
-	if guid, err := uuid.FromBytes(fv.FileSystemGUID[:]); err == nil {
-		guidString = guid.String()
-		guidName, ok = FirmwareVolumeGUIDs[guidString]
-		if !ok {
-			guidName = "Unknown"
-		}
-	} else {
-		guidString = "<invalid GUID>"
-		guidName = "Unknown"
-	}
 	return fmt.Sprintf("FirmwareVolume{\n"+
 		"    FileSystemGUID=%s (%v)\n"+
 		"    Length=%v\n"+
@@ -84,7 +74,7 @@ func (fv FirmwareVolume) Summary() string {
 		"    Revision=%v\n"+
 		"    Blocks=%v\n"+
 		"}",
-		guidString, guidName,
+		fv.guidString, fv.guidName,
 		fv.Length, fv.Signature, fv.AttrMask,
 		fv.HeaderLen, fv.Checksum, fv.Revision,
 		fv.Blocks,
@@ -93,7 +83,7 @@ func (fv FirmwareVolume) Summary() string {
 
 // FindFirmwareVolumeOffset searches for a firmware volume signature, "_FVH"
 // using 8-byte alignment. If found, returns the offset from the start of the
-// firmware volume, otherwise returns -1.
+// bios region, otherwise returns -1.
 func FindFirmwareVolumeOffset(data []byte) int64 {
 	if len(data) < 32 {
 		return -1
@@ -113,13 +103,15 @@ func FindFirmwareVolumeOffset(data []byte) int64 {
 // NewFirmwareVolume parses a sequence of bytes and returns a FirmwareVolume
 // object, if a valid one is passed, or an error
 func NewFirmwareVolume(data []byte) (*FirmwareVolume, error) {
+	var fv FirmwareVolume
+	var ok bool
+
 	if len(data) < FirmwareVolumeMinSize {
 		return nil, fmt.Errorf("Firmware Volume size too small: expected %v bytes, got %v",
 			FirmwareVolumeMinSize,
 			len(data),
 		)
 	}
-	var fv FirmwareVolume
 	reader := bytes.NewReader(data)
 	if err := binary.Read(reader, binary.LittleEndian, &fv.FirmwareVolumeFixedHeader); err != nil {
 		return nil, err
@@ -138,5 +130,16 @@ func NewFirmwareVolume(data []byte) (*FirmwareVolume, error) {
 		blocks = append(blocks, block)
 	}
 	fv.Blocks = blocks
+
+	if guid, err := uuid.FromBytes(fv.FileSystemGUID[:]); err == nil {
+		fv.guidString = guid.String()
+		fv.guidName, ok = FirmwareVolumeGUIDs[fv.guidString]
+		if !ok {
+			fv.guidName = "Unknown"
+		}
+	} else {
+		fv.guidString = "<invalid GUID>"
+		fv.guidName = "Unknown"
+	}
 	return &fv, nil
 }
--- a/uefi/flash.go
+++ b/uefi/flash.go
@@ -142,8 +142,8 @@ func NewFlashImage(buf []byte) (*FlashImage, error) {
 	flash.Master = *master
 
 	// BIOS region
-	biosBase := uint32(uint32(flash.Region.BiosBase) * 0x1000)
-	biosSize := uint32(computeRegionSize(flash.Region.BiosBase, flash.Region.BiosLimit))
+	biosBase := uint32(flash.Region.BIOS.Base) * 0x1000
+	biosSize := computeRegionSize(flash.Region.BIOS.Base, flash.Region.BIOS.Limit)
 	br, err := NewBiosRegion(buf[biosBase : biosBase+biosSize])
 	if err != nil {
 		return nil, err
--- a/uefi/flashmastersection.go
+++ b/uefi/flashmastersection.go
@@ -9,23 +9,29 @@ import (
 // FlashMasterSectionSize is the size in bytes of the FlashMaster section
 const FlashMasterSectionSize = 12
 
+// RegionPermissions holds the read/write permissions for other regions.
+type RegionPermissions struct {
+	ID    uint16
+	Read  uint8
+	Write uint8
+}
+
+func (r RegionPermissions) String() string {
+	return fmt.Sprintf("RegionPermissions{ID=%v, Read=0x%x, Write=0x%x}",
+		r.ID, r.Read, r.Write)
+}
+
 // FlashMasterSection holds all the IDs and read/write permissions for other regions
 // This controls whether the bios region can read/write to the ME for example.
 type FlashMasterSection struct {
-	BiosID    uint16
-	BiosRead  uint8
-	BiosWrite uint8
-	MeID      uint16
-	MeRead    uint8
-	MeWrite   uint8
-	GbeID     uint16
-	GbeRead   uint8
-	GbeWrite  uint8
+	BIOS RegionPermissions
+	ME   RegionPermissions
+	GBE  RegionPermissions
 }
 
 func (m FlashMasterSection) String() string {
-	return fmt.Sprintf("FlashMasterSection{BiosID=%v, MeID=%v, GbeID=%v}",
-		m.BiosID, m.MeID, m.GbeID)
+	return fmt.Sprintf("FlashMasterSection{Bios %v, Me %v, Gbe %v}",
+		m.BIOS, m.ME, m.GBE)
 }
 
 // Summary prints a multi-line description of the FlashMasterSection
@@ -41,9 +47,9 @@ func (m FlashMasterSection) Summary() string {
 		"    GbeRead=%v\n"+
 		"    GbeWrite=%v\n"+
 		"}",
-		m.BiosID, m.BiosRead, m.BiosWrite,
-		m.MeID, m.MeRead, m.MeWrite,
-		m.GbeID, m.GbeRead, m.GbeWrite,
+		m.BIOS.ID, m.BIOS.Read, m.BIOS.Write,
+		m.ME.ID, m.ME.Read, m.ME.Write,
+		m.GBE.ID, m.GBE.Read, m.GBE.Write,
 	)
 }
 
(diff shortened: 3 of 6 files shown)


uefi/firmwarevolume.go
```diff
@@ -58,12 +59,13 @@ type FirmwareVolume struct {
...
 }
 
-// Summary prints a multi-line representation of a FirmwareVolume object
-func (fv FirmwareVolume) Summary() string {
-	var (
-		guidName, guidString string
-		ok                   bool
-	)
+// This is required to avoid recursion while marshalling json
+type jsonFirmwareVolume FirmwareVolume
```
hugelgupf: unused?

uefi/firmwarevolume.go
```diff
@@ -58,12 +59,13 @@ type FirmwareVolume struct {
...
+// MarshalJSON marshals the Firmware Volume into JSON
+func (fv FirmwareVolume) MarshalJSON() ([]byte, error) {
+	var guidName, guidString string
+	var ok bool
+
 	if guid, err := uuid.FromBytes(fv.FileSystemGUID[:]); err == nil {
```
hugelgupf: Why not just use the existing FileSystemGUID field? This seems needlessly complicated?

uefi/firmwarevolume.go
```diff
@@ -74,26 +76,22 @@ func (fv FirmwareVolume) Summary() string {
...
-		"    Checksum=0x%04x\n"+
-		"    Revision=%v\n"+
-		"    Blocks=%v\n"+
-		"}",
-		guidString, guidName,
-		fv.Length, fv.Signature, fv.AttrMask,
-		fv.HeaderLen, fv.Checksum, fv.Revision,
-		fv.Blocks,
-	)
+	return json.Marshal(struct {
```
hugelgupf: type-declare this struct, so you can also unmarshal

hugelgupf: why not add stuff like the name to FirmwareVolume?

uefi/flash.go
```diff
@@ -26,6 +27,18 @@ type FlashImage struct {
 	BiosRegion *BiosRegion
 }
 
+type jsonFlashImage FlashImage
+
+func (f FlashImage) MarshalJSON() ([]byte, error) {
+	return json.Marshal(struct {
+		Size int
```
hugelgupf: Why the size? Unnecessary afaict

hugelgupf: just json.Marshal(f) should work?

uefi/flashregionsection.go
```diff
@@ -12,14 +13,26 @@ const FlashRegionSectionSize = 36
...
 }
 
+type jsonFlashRegionSection FlashRegionSection
+// MarshalJSON marshals the FlashRegionSection into JSON
+func (f FlashRegionSection) MarshalJSON() ([]byte, error) {
+	return json.Marshal(struct {
+		AvailableRegions string `json:"Regions"`
+		jsonFlashRegionSection
+	}{
+		AvailableRegions : strings.Join(f.AvailableRegions(), ","),
```
hugelgupf: What's the point? You're just encoding the data twice in the JSON through both the Regions field and BiosLimit.

hugelgupf: I think if you're doing this for readability, all these things should just have a fmt.Stringer for the human-readable stuff (e.g. fmt.Sprintf("FlashRegionSecton{%v, AvailableRegions=%s}", f, f.AvailableRegions())). Let's not do that in JSON.

uefi/flashregionsection.go
```diff
@@ -12,14 +13,26 @@ const FlashRegionSectionSize = 36
 
 // FlashRegionSection holds the metadata of all the different flash regions like PDR, Gbe and the Bios region.
 type FlashRegionSection struct {
-	Reserved            uint16
+	Reserved            uint16 `json:"-"`
 	FlashBlockEraseSize uint16
 	BiosBase, BiosLimit uint16
```
hugelgupf: How about something like

```go
type Region struct {
  Base  uint16
  Limit uint16
}

func (r Region) Available() bool {
  return r.Limit != 0
}

func (r Region) String() string {
  return fmt.Sprintf("[%d, %d)", r.Base, r.Limit)
}

type FlashRegionSection struct {
  BIOS Region
  ME   Region
  GBE  Region
  PDR  Region
}
```

hugelgupf: go nit: capitalize full thing (BIOS, ME, GBE)

uefi/flashregionsection.go
hugelgupf: we should separate human-readable strings from machine encoding.

uefi/firmwarevolume.go
GanShun: I wanted to make this human readable.

GanShun: I should also keep the FileSystemGUID variable, instead of not marshalling it.

uefi/flash.go
GanShun: Size is nice for looking at the summary, I can get rid of it. If I don't need the size then I don't need all this stuff.

uefi/firmwarevolume.go
GanShun: oh nice catch. was a leftover.

uefi/flashregionsection.go
GanShun: good idea, I'll change that.

uefi/flashregionsection.go
GanShun: That's a fair point but I didn't want to have separate parsing rules for both of them. I was hoping we could just have the same json for both humans and machine encoding.

Should I then keep the original summary stuff?

uefi/firmwarevolume.go
GanShun: I could do that. I should differentiate between what's actually in the binary and what isn't now.

uefi/flashregionsection.go
hugelgupf: Maybe keep some of them (those where you're unhappy with %+v), and make them fmt.Stringer implementations instead of our own name.

hugelgupf: I guess the point I want to make is this: I'd like to see tests for the machine encoding with marshaling and unmarshaling. If the marshaling and unmarshaling is just json.Marshal/json.Unmarshal that pretty well covers it, because those are well-tested. But if we're to implement our own MarshalJSON and UnmarshalJSON methods, we have to test them. 

We don't have to really test the human-readable stuff. So if you separate it such that JSON stuff is *just* json.Marshal and json.Unmarshal, and human-readable stuff is whatever, no tests necessary.

uefi/flashregionsection.go
GanShun: Done

GanShun:
I've kept the summary functions, but parse dumps JSON. We can figure out another command if we want more human readable stuff, that can replace the original Parse functionality

uefi/flashregionsection.go
```diff
@@ -10,29 +10,43 @@ import (
 // FlashRegionSectionSize is the size of the Region descriptor. It is made up by 18 fields, each 16-bits large.
 const FlashRegionSectionSize = 36
 
+//
```
hugelgupf: finish comment?

uefi/flashregionsection.go
```diff
@@ -54,10 +68,10 @@ func (f FlashRegionSection) Summary() string {
 		"    PdrBase=%v (size: %v)\n"+
 		"}",
 		strings.Join(f.AvailableRegions(), ","),
-		f.BiosBase, f.BiosLimit,
-		f.MeBase, f.MeLimit,
-		f.GbeBase, f.GbeLimit,
-		f.PdrBase, f.PdrLimit,
+		f.BIOS.Base, f.BIOS.Limit,
```
hugelgupf: Since Region has a fmt.Stringer, maybe just fmt.Sprintf(..."BIOS=%v"..., ...f.BIOS...)?

uefi/flashmastersection.go
```diff
@@ -41,9 +42,9 @@ func (m FlashMasterSection) Summary() string {
 		"    GbeRead=%v\n"+
 		"    GbeWrite=%v\n"+
 		"}",
-		m.BiosID, m.BiosRead, m.BiosWrite,
-		m.MeID, m.MeRead, m.MeWrite,
-		m.GbeID, m.GbeRead, m.GbeWrite,
+		m.BIOS.ID, m.BIOS.Read, m.BIOS.Write,
```
hugelgupf: same here - RegionPermissions could have a stringer?

hugelgupf: Actually, in general - can we skip the multi-line summaries? We should have compact fmt.Stringers, and if you want multi-line summaries like this, you should use a deep pretty-printing library like https://godoc.org/github.com/kr/pretty or https://github.com/davecgh/go-spew - reflection does a much better job than people here

Anyway, this is all optional, your judgment

GanShun: I agree, but I don't wanna remove the original summary stuff for this pr. we can do that in another one.

uefi/flashregionsection.go
GanShun: Done

uefi/flashregionsection.go
GanShun: Lets do this another pr

uefi/firmwarevolume.go
```diff
@@ -37,16 +37,16 @@ type Block struct {
 // FirmwareVolumeFixedHeader contains the fixed fields of a firmware volume
 // header
 type FirmwareVolumeFixedHeader struct {
-	Zeros          [16]uint8
+	Zeros          [16]uint8 `json:"-"`
```
rminnich: just fyi, the other way to do this I believe is just name it _. that's how I've done it in the past, then you don't the json gorp.
I prefer to just use _ in most cases because in one struct I run out of names like
unused1
unused2
etc. And the name _ makes it very clear you don't care.

uefi/firmwarevolume.go
```diff
@@ -37,16 +37,16 @@ type Block struct {
...
 	FileSystemGUID [16]uint8
 	Length         uint64
 	Signature      uint32
 	AttrMask       uint8
 	HeaderLen      uint16
 	Checksum       uint16
-	Reserved       [3]uint8
+	Reserved       [3]uint8  `json:"-"`
```
rminnich: and, fwiw, intel often calls things unused, and reserved, interchangeably, then uses them anyway. See: 386 PTEs.

uefi/flash.go
```diff
@@ -142,8 +142,8 @@ func NewFlashImage(buf []byte) (*FlashImage, error) {
 	flash.Master = *master
 
 	// BIOS region
-	biosBase := uint32(uint32(flash.Region.BiosBase) * 0x1000)
-	biosSize := uint32(computeRegionSize(flash.Region.BiosBase, flash.Region.BiosLimit))
+	biosBase := uint32(uint32(flash.Region.BIOS.Base) * 0x1000)
```
rminnich: why the double cast? was it really needed?

uefi/flashmastersection.go
```diff
@@ -9,23 +9,24 @@ import (
 // FlashMasterSectionSize is the size in bytes of the FlashMaster section
 const FlashMasterSectionSize = 12
 
+// RegionPermissions holds the
```
rminnich: holds the what? the what? We don't have much time here, I've got to know!

uefi/flashregionsection.go
```diff
@@ -10,29 +10,44 @@ import (
 // FlashRegionSectionSize is the size of the Region descriptor. It is made up by 18 fields, each 16-bits large.
 const FlashRegionSectionSize = 36
 
+// Region contains a the start and end of a region in flash. This can be a BIOS, ME, PDR or GBE region.
+type Region struct {
```
rminnich: uint16? really? If it's a block number or something you might want to indicate in the comment, OR make a type to use. uint16 is confusing to me.

uefi/flashregionsection.go
```diff
@@ -10,29 +10,44 @@ import (
...
+	Base  uint16
+	Limit uint16
+}
+
+// Available checks to see if a region is valid (has nonzero limit)
```
rminnich: I would take out the (has nonzero limit) bit because this function could end up doing more later.

uefi/flashregionsection.go
```diff
@@ -10,29 +10,44 @@ import (
...
+func (r *Region) Available() bool {
+	return r.Limit > 0
+}
+
+func (r *Region) String() string {
```
rminnich: me, personally, I would prefer hex but.

uefi/flashregionsection.go
```diff
@@ -10,29 +10,44 @@ import (
...
+	return fmt.Sprintf("[%d, %d)", r.Base, r.Limit)
+}
+
 // FlashRegionSection holds the metadata of all the different flash regions like PDR, Gbe and the Bios region.
 type FlashRegionSection struct {
-	Reserved            uint16
+	Reserved            uint16 `json:"-"`
```
rminnich: consider _

rminnich: https://play.golang.org/p/yco3mh2BUVZ

uefi/flashmastersection.go
GanShun: I don't know why I'm dropping all these comments dammit.

uefi/flashregionsection.go
GanShun: ok

uefi/firmwarevolume.go
rminnich: and it works.
https://play.golang.org/p/kttTW44tA3I

uefi/firmwarevolume.go
GanShun: If I remember correctly, I do see some things in this field when I parse an actual rom image, but we ignore them right now

uefi/flashregionsection.go
GanShun: done

uefi/flash.go
GanShun: removed and removed the one below

uefi/firmwarevolume.go
GanShun: done

uefi/flashregionsection.go
GanShun: done. This isn't actually used right now, json exports as decimal anyway.

uefi/flashregionsection.go
GanShun: done

uefi/firmwarevolume.go
```diff
@@ -37,16 +37,16 @@ type Block struct {
...
-	Zeros          [16]uint8
+	_              [16]uint8
 	FileSystemGUID [16]uint8
 	Length         uint64
 	Signature      uint32
 	AttrMask       uint8
 	HeaderLen      uint16
 	Checksum       uint16
-	Reserved       [3]uint8
+	Reserved       [3]uint8 `json:"-"`
```
hugelgupf: why no _?

uefi/firmwarevolume.go
```diff
@@ -138,5 +130,16 @@ func NewFirmwareVolume(data []byte) (*FirmwareVolume, error) {
 		blocks = append(blocks, block)
 	}
 	fv.Blocks = blocks
+
+	if guid, err := uuid.FromBytes(fv.FileSystemGUID[:]); err == nil {
+		fv.guidString = guid.String()
+		fv.guidName, ok = FirmwareVolumeGUIDs[fv.guidString]
```
hugelgupf: declare var ok bool here

uefi/flashregionsection.go
```diff
@@ -10,29 +10,46 @@ import (
...
+	Limit uint16
+}
+
+// Available checks to see if a region is valid
+func (r *Region) Available() bool {
+	return r.Limit > 0
+}
+
+func (r *Region) String() string {
+	return fmt.Sprintf("[0x%x, 0x%x)", r.Base, r.Limit)
```
hugelgupf: make that `"[%#x, %#x)"`

hugelgupf:
Eh fuck it actually. I'm just going to merge it.

GanShun:
Will bring those nits to the next pr

PR #5, opened 2018-06-07
Utk support for subcommands

GanShun:
This allows us to do stuff like utk parse <rom> or utk extract <rom> <location> in the future.

### DIFF
--- a/utk/utk.go
+++ b/utk/utk.go
@@ -1,40 +1,79 @@
 package main
 
 import (
+	"context"
+	"encoding/json"
 	"flag"
 	"fmt"
 	"io/ioutil"
 	"log"
 	"os"
-	"encoding/json"
 
+	"github.com/google/subcommands"
 	"github.com/linuxboot/fiano/uefi"
 )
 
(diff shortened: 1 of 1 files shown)


utk/utk.go
```diff
@@ -6,35 +6,73 @@ import (
...
+}
+
+func (*parseCmd) Usage() string {
+	return "parse <path-to-rom-file>\n"
+}
+
+func (*parseCmd) SetFlags(_ *flag.FlagSet) {}
+
+func (*parseCmd) Execute(_ context.Context, f *flag.FlagSet, _ ...interface{}) subcommands.ExitStatus {
+	if len(f.Args()) == 0 {
```
rminnich: This is all fine, one thought: I once had it suggested to me that if you use f.Args() more than once, you should just do args := f.Args() since the call cost is nontrivial. I don't necessarily agree with this, but thought I'd mention it.

rminnich: chris, in case you doing me on this, it was a comment from Adin :-)

PR #8, opened 2018-06-11
Removing Summary

GanShun:
Remove Indent functions as well.


insomniacslk:
Remind me what are we going to use instead of Summary?

GanShun:
Sorry, Ron merged this already. Right now we have the json output that's mildly readable. Also #4 chris commented this:

"Actually, in general - can we skip the multi-line summaries? We should have compact fmt.Stringers, and if you want multi-line summaries like this, you should use a deep pretty-printing library like https://godoc.org/github.com/kr/pretty or https://github.com/davecgh/go-spew - reflection does a much better job than people here

Anyway, this is all optional, your judgment"

That's what I'm leaning towards using too.

However I can revert this if we feel strongly about it.

insomniacslk:
no, absolutely, this is all good. I was just asking what was the plan as I missed some context during my holiday break. Thanks!

PR #10, opened 2018-06-12
Extract

GanShun:
Early simplistic extract command

### DIFF
--- a/uefi/biosregion.go
+++ b/uefi/biosregion.go
@@ -1,5 +1,9 @@
 package uefi
 
+import (
+	"os"
+)
+
 // BIOSRegion represents the Bios Region in the firmware.
 // It holds all the FVs as well as padding
 // TODO(ganshun): handle padding
@@ -35,4 +39,26 @@ func NewBIOSRegion(buf []byte, r *Region) (*BIOSRegion, error) {
 		// search for FVs (i.e. offset + fv.size)
 	}
 	return &br, nil
+}
+
+// Extract extracts the Bios Region to the directory passed in.
+func (br *BIOSRegion) Extract(dirPath string) error {
+	// Create the directory if it doesn't exist
+	err := os.MkdirAll(dirPath, 0755)
+	if err != nil {
+		return err
+	}
+
+	// Dump the binary.
+	br.ExtractPath = dirPath + "/biosregion.bin"
+	binFile, err := os.OpenFile(br.ExtractPath, os.O_RDWR|os.O_CREATE, 0755)
+	if err != nil {
+		return err
+	}
+	defer binFile.Close()
+	_, err = binFile.Write(br.buf)
+	if err != nil {
+		return err
+	}
+	return nil
 }
--- a/uefi/biosregion.go
+++ b/uefi/biosregion.go
@@ -1,9 +1,5 @@
 package uefi
 
-import (
-	"os"
-)
-
 // BIOSRegion represents the Bios Region in the firmware.
 // It holds all the FVs as well as padding
 // TODO(ganshun): handle padding
@@ -43,22 +39,8 @@ func NewBIOSRegion(buf []byte, r *Region) (*BIOSRegion, error) {
 
 // Extract extracts the Bios Region to the directory passed in.
 func (br *BIOSRegion) Extract(dirPath string) error {
-	// Create the directory if it doesn't exist
-	err := os.MkdirAll(dirPath, 0755)
-	if err != nil {
-		return err
-	}
-
-	// Dump the binary.
-	br.ExtractPath = dirPath + "/biosregion.bin"
-	binFile, err := os.OpenFile(br.ExtractPath, os.O_RDWR|os.O_CREATE, 0755)
-	if err != nil {
-		return err
-	}
-	defer binFile.Close()
-	_, err = binFile.Write(br.buf)
-	if err != nil {
-		return err
-	}
-	return nil
+	// We just dump the binary for now
+	var err error
+	br.ExtractPath, err = ExtractBinary(br.buf, dirPath, "biosregion.bin")
+	return err
 }
--- a/uefi/flash.go
+++ b/uefi/flash.go
@@ -3,7 +3,10 @@ package uefi
 import (
 	"bytes"
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
+	"os"
+	"path/filepath"
 )
 
 // FlashSignature is the sequence of bytes that a Flash image is expected to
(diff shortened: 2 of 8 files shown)


uefi/flash.go
```diff
@@ -84,6 +109,88 @@ func (f *FlashImage) Validate() []error {
 	return errors
 }
 
+// Extract extracts the flash image to the directory passed in.
+func (f *FlashImage) Extract(dirPath string) error {
```
insomniacslk: same feedback as above about collapsing error handling lines and using `os.Create`

uefi/flash.go
```diff
@@ -84,6 +109,88 @@ func (f *FlashImage) Validate() []error {
...
+	absDirPath, err := filepath.Abs(dirPath)
+	if err != nil {
+		return err
+	}
+	// Create the directory if it doesn't exist
+	err = os.MkdirAll(absDirPath, 0755)
```
insomniacslk: if I am not mistaken, the various `BIOSRegion`, `MERegion` etc implement the `Firmware` interface (`Summary()` and `Validate()`). In this case, what about adding an `Extract` method to the `Firmware` interface? It would be more consistent, and here you would have to call `if err := region.Extract(); err != nil { ... }` for each region. The `Extract` method would also have to check whether the region is not `nil`

uefi/biosregion.go
```diff
@@ -54,3 +55,25 @@ func NewBIOSRegion(buf []byte, r *Region) (*BIOSRegion, error) {
 	}
 	return &br, nil
 }
+
+// Extract extracts the Bios Region to the directory passed in.
+func (br *BIOSRegion) Extract(dirPath string) error {
+	// Create the directory if it doesn't exist
+	err := os.MkdirAll(dirPath, 0755)
+	if err != nil {
```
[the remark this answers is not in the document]
GanShun: doing so.

uefi/biosregion.go
```diff
@@ -54,3 +55,25 @@ func NewBIOSRegion(buf []byte, r *Region) (*BIOSRegion, error) {
...
+		return err
+	}
+
+	// Dump the binary.
+	br.ExtractPath = dirPath + "/biosregion.bin"
```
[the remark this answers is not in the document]
GanShun: I think it should be a directory, because extracting this bios region may result in multiple files. For example if we want to extract the firmware volumes later, I don't want to have to pass in multiple filenames or have to remove the "biosregion.bin" that was passed in to call the sub directories. 

Good point on the path.join though. I'll change that.

uefi/biosregion.go
```diff
@@ -54,3 +55,25 @@ func NewBIOSRegion(buf []byte, r *Region) (*BIOSRegion, error) {
...
+	binFile, err := os.OpenFile(br.ExtractPath, os.O_RDWR|os.O_CREATE, 0755)
```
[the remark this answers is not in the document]
GanShun: Will do. I was just lazy and wanted to do it in one instruction =p

uefi/flash.go
GanShun: It's actually been added to the Firmware interface already, but I think they're all different still. I could move the nil check into the extract functions for each of the regions, but I'm not sure what that gains us.

uefi/flash.go
GanShun: will do

uefi/flash.go
```diff
@@ -31,6 +34,28 @@ type FlashDescriptor struct {
 	ExtractPath string
 }
 
+// Extract extracts the flash descriptor region to the directory passed in.
+func (fd *FlashDescriptor) Extract(dirPath string) error {
```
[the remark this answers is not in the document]
GanShun: I could refactor this, but the intention is to add more differing functionality later. For example in BiosRegion's extract, we'll want to extract the FVs. I'll try to move common functionality into a shared helper function though. I think you're right and we can at least share the binary dumping portions of this extract function

uefi/biosregion.go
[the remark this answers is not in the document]
hugelgupf: everything honors umask, including open(2). just do the open.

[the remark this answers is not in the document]
GanShun: got it.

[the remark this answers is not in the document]
hugelgupf: make it not executable though. 0666.

uefi/biosregion.go
```diff
@@ -54,3 +55,25 @@ func NewBIOSRegion(buf []byte, r *Region) (*BIOSRegion, error) {
...
+	if err != nil {
+		return err
+	}
+	defer binFile.Close()
+	_, err = binFile.Write(br.buf)
+	if err != nil {
```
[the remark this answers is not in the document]
GanShun: done

uefi/biosregion.go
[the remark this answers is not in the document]
GanShun: done

uefi/flash.go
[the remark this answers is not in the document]
GanShun: done. it's much nicer now.

uefi/flash.go
GanShun: collapsed the error handling.

uefi/uefi.go
```diff
@@ -25,3 +28,24 @@ func Parse(buf []byte) (Firmware, error) {
...
+// It returns the filepath of the binary, and an error if it exists.
+// This is meant as a helper function for other Extract functions.
+func ExtractBinary(buf []byte, dirPath string, filename string) (string, error) {
+	// Create the directory if it doesn't exist
+	if err := os.MkdirAll(dirPath, 0755); err != nil {
+		return "", err
+	}
+
+	// Dump the binary.
+	fp := filepath.Join(dirPath, filename)
```
hugelgupf: https://golang.org/pkg/io/ioutil/#WriteFile

uefi/uefi.go
```diff
@@ -25,3 +28,24 @@ func Parse(buf []byte) (Firmware, error) {
...
+	binFile, err := os.OpenFile(fp, os.O_RDWR|os.O_CREATE, 0666)
+	if err != nil {
+		return fp, err
```
hugelgupf: return "", err

uefi/uefi.go
```diff
@@ -25,3 +28,24 @@ func Parse(buf []byte) (Firmware, error) {
...
+	}
+	defer binFile.Close()
+	_, err = binFile.Write(buf)
```
hugelgupf: if err != nil {
  return "", err
}

uefi/flash.go
```diff
@@ -90,6 +101,71 @@ func (f *FlashImage) Validate() []error {
...
+	if err != nil {
+		return err
+	}
+	defer summary.Close()
+	b, err := json.MarshalIndent(f, "", "    ")
+	if err != nil {
+		return err
+	}
+	_, err = summary.Write(b)
+	if err != nil {
```
hugelgupf: return err

uefi/flash.go
```diff
@@ -32,6 +35,14 @@ type FlashDescriptor struct {
 	ExtractPath string
 }
 
+// Extract extracts the flash descriptor region to the directory passed in.
+func (fd *FlashDescriptor) Extract(dirPath string) error {
```
hugelgupf: s/Extract/WriteContentTo?

rminnich: me like extract.shorter.

uefi/flash.go
```diff
@@ -90,6 +101,61 @@ func (f *FlashImage) Validate() []error {
...
+		return err
+	}
+	// Dump the binary
+	f.ExtractPath, err = ExtractBinary(f.buf, absDirPath, "flash.rom")
+	if err != nil {
+		return err
+	}
+
+	// Extract IFD
+	if err = f.IFD.Extract(filepath.Join(absDirPath, "ifd")); err != nil {
```
rminnich: it sure seems like each of these Extract functions has the same type signature. Were you to make Extract an interface, then you could have a nice little struct array with pointers to functions that implement the interface, which these all seem to do, and it would be just a bit tidier (and more friendly to a Fuse implementation later).

rminnich: type Extracter interface {
Extract(dir, base string) error
}

GanShun: Actually Extract is a function in the Firmware Interface shown in uefi.go

PR #11, opened 2018-06-13
Unit tests to improve coverage of UEFI library files.

subhav:


uefi/region.go
```diff
@@ -17,13 +17,13 @@ const (
...
-	Base  uint16
-	Limit uint16
+	Base  uint16 // Index of first 4k block
+	Limit uint16 // Index of last block
 }
 
 // Valid checks to see if a region is valid
 func (r *Region) Valid() bool {
-	return r.Limit > 0
+	return r.Limit > 0 && r.Limit >= r.Base
```
GanShun: Nice. Is this from the spec? Or was it something you thought of? Either way this is good, I just wanted to know.

uefi/region_test.go
```diff
@@ -0,0 +1,61 @@
...
+		base:  0x1000,
+		end:   0x2000,
+	},
+	{
+		in:    &Region{100, 200},
+		valid: true,
+		base:  0x64000,
+		end:   0xC9000,
+	},
+}
```
GanShun: Good stuff, one suggestion: add a test where the limit is right at the largest possible value.

uefi/region_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+func TestRegion_Valid(t *testing.T) {
+	for _, tc := range regionTestcases {
+		if out := (*Region).Valid(tc.in); out != tc.valid {
```
GanShun: Can't you do tc.in.Valid() here?

GanShun: Likewise for all other instances of (*Region).Valid

uefi/region_test.go
```diff
@@ -0,0 +1,61 @@
+package uefi
+
+import "testing"
+
+var regionTestcases = [...]struct {
+	in    *Region
```
GanShun: Is there a reason you made it a region pointer instead of just putting the region struct in?

uefi/region.go
subhav: No, just something I thought of.

Is this Valid check supposed comply with a specific definition in the spec? If it is, we might want to make that clear in the function comment.

GanShun: no clue. you can look it up and tell me =P

uefi/region_test.go
subhav: I think using `(*Region).Valid(%#v)` makes sense in the error strings. This way I'm consistent with that.

Let me know if you still want it to change.

GanShun: I think this is confusing though. In the error strings I'd much rather something like:

t.Errorf("Region %v.Valid() = %v, expected %v", *tc.in, out, tc.valid)

uefi/region_test.go
subhav: The pointer is the actual input for the methods being tested. Apart from that, no.

Should I change this?

uefi/region_test.go
subhav: Done.

uefi/region_test.go
GanShun: no reason to make it a pointer, just embed the struct imo

uefi/region.go
insomniacslk: it wasn't thought with the specs in mind. Most of the original work on this library is coming from a mix of specs, UEFITool and uefi-firmware-parser

PR #12, opened 2018-06-15
Use Firmware interface for extraction and Add Validate functions for Regions.

GanShun:


uefi/flash.go
```diff
@@ -57,7 +64,8 @@ type FlashImage struct {
 	PD   *PDRegion
 
 	// Metadata for extraction and recovery
-	ExtractPath string
+	ExtractPath    string
+	extractRegions []Firmware
```
insomniacslk: could this just be `regions` or `regionList` ?

uefi/gberegion.go
```diff
@@ -18,6 +22,19 @@ func NewGBERegion(buf []byte, r *Region) (*GBERegion, error) {
 	return &gbe, nil
 }
 
+// Validate Region
+func (gbe *GBERegion) Validate() []error {
+	// TODO: Add more verification if needed.
+	errs := make([]error, 0)
+	if gbe.Position == nil {
+		errs = append(errs, fmt.Errorf("GBERegion position is nil"))
```
insomniacslk: nit: I have no strong opinions and don't know if there's any significant difference, but since there's no format this could be `errors.New` instead of `fmt.Errorf` (feel free to ignore)

uefi/gberegion.go
```diff
@@ -18,6 +22,19 @@ func NewGBERegion(buf []byte, r *Region) (*GBERegion, error) {
...
+	}
+	if !gbe.Position.Valid() {
+		errs = append(errs, fmt.Errorf("GBERegion is not valid, region was %v", gbe.Position))
```
insomniacslk: if I'm not wrong you will have the pointer of `gbe.Position` in this format string. Is this what you want?

(Same question for the other regions below)

GanShun: Nice catch! Thanks!

uefi/gberegion.go
GanShun: done!

uefi/flash.go
GanShun: done!

GanShun:
Adjusted title. I named it that because I only added the validate functions so the regions would implement the firmware interface =P

insomniacslk:
You are right. Well, now I made you change the title already, too late. :-P
Merging this since it looks good and no other feedback came in

PR #17, opened 2018-06-22
Clean up UUID code

GanShun:
Clean up parsing and String functions, as well as use UUID type in
headers rather than maintaining a separate guid.

This needs to be rebased on top of #15, I basically need to make the same change for firmwarefile.go. Do not merge, but please review.

### DIFF
--- a/uuid/uuid.go
+++ b/uuid/uuid.go
@@ -1,74 +1,74 @@
-package uefi
+package uuid
 
 // This package implements the mixed-endian UUID as implemented by Microsoft.
 
 import (
 	"encoding/hex"
-	"errors"
+	"fmt"
+	"strings"
 )
-
-var ErrInvalidUUID = errors.New("Invalid UUID")
 
 const (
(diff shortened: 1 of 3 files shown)


uuid/uuid.go
```diff
@@ -1,74 +1,74 @@
...
+)
+
+var (
+	fields = [...]int{4, 2, 2, 2, 1, 1, 1, 1, 1, 1}
 )
 
-type UUID struct {
-	Data []byte
+// UUID represents a unique identifier.
+type UUID [Size]byte
```
rminnich: I'm wondering why you can't use github.com/google/uuid. That's what I did in the u-root gpt package. You're doing a lot of work here that the uuid package already does ...

GanShun: that doesn't do endianess. IIRC it assumes big endian

insomniacslk: that's exactly why I originally reimplemented it instead of using an existing package

PR #22, opened 2018-06-24
Json UUID

GanShun:
We need this to be able to even read the json file. I know we want to keep human readable stuff as separate as possible, but no one is going to be able to modify the json otherwise.

### DIFF
--- a/uuid/uuid.go
+++ b/uuid/uuid.go
@@ -4,6 +4,7 @@ package uuid
 
 import (
 	"encoding/hex"
+	"encoding/json"
 	"fmt"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


uuid/uuid.go
```diff
@@ -72,3 +73,25 @@ func (u UUID) String() string {
...
+	j := make(map[string]string)
+
+	if err := json.Unmarshal(b, &j); err != nil {
+		return err
+	}
+	g, err := Parse(j["UUID"])
+	if err != nil {
+		return err
+	}
+	copy(u[:], g[:])
```
insomniacslk: should you check that `len(g) == Size` before copying?

GanShun: both of them are UUIDs, the length is fixed since it's an array, not a slice.

insomniacslk: you're right, I didn't realize that

PR #23, opened 2018-06-25
Add Validate for Firmware Files.

GanShun:

### DIFF
--- a/uefi/firmwarefile.go
+++ b/uefi/firmwarefile.go
@@ -70,11 +70,11 @@ func (a fileAttr) isLarge() bool {
 }
 
 // Checks if we need to checksum the file body
-func (a fileAttr) checksum() bool {
+func (a fileAttr) hasChecksum() bool {
 	return a&0x40 != 0
 }
 
-func (f *FirmwareFile) readSize() uint64 {
+func (f *FirmwareFile) size() uint64 {
 	return uint64(f.Header.Size[2])<<16 |
 		uint64(f.Header.Size[1])<<8 | uint64(f.Header.Size[0])
 }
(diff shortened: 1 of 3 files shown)


uefi/firmwarefile.go
```diff
@@ -41,16 +41,42 @@ const (
...
+	State      uint8
+}
+
+// Checks if the large file attribute is set
+func (a fileAttr) isLarge() bool {
+	return a&0x01 != 0
+}
+
+// Checks if we need to checksum the file body
+func (a fileAttr) checksum() bool {
```
insomniacslk: the name `checksum` is a bit counter-intuitive, it hints to a method that will return a checksum rather than a method that will return whether the checksum is set and has to be verified. What about renaming it to `hasChecksum` or similar?

uefi/firmwarefile.go
```diff
@@ -41,16 +41,42 @@ const (
...
+	return a&0x40 != 0
+}
+
+func (f *FirmwareFile) readSize() uint64 {
```
insomniacslk: this method's name could probably just be `size`. Also, worth making it public? Sounds like something we may want to expose to the users

uefi/firmwarefile.go
```diff
@@ -69,6 +95,79 @@ type FirmwareFile struct {
...
+	buflen := uint64(len(f.buf))
+	blankSize := [3]byte{0xFF, 0xFF, 0xFF}
+	if buflen < FileHeaderMinLength {
+		errs = append(errs, fmt.Errorf("length too small!, buffer is only %#x bytes long", buflen))
+		return errs
+	}
+
+	// Size Checks
+	fh := &f.Header
+	if fh.Size == blankSize && buflen < FileHeaderExtMinLength {
```
insomniacslk: nit: you could group all the checks when the fh is blank under a single `if fh.Size == blankSize`, then in the body check for `buflen < FIleHeaderExtMinLength` and `!fh.Attributes.isLarge()` separately. Then an `else` for the other case.
It's more for readability (and to be future-proof in case of changes) but feel free to ignore

uefi/firmwarefile.go
```diff
@@ -69,6 +95,79 @@ type FirmwareFile struct {
...
+	// Header Checksums
+	var headerSize int
+	if fh.Attributes.isLarge() {
+		headerSize = FileHeaderExtMinLength
+	} else {
+		headerSize = FileHeaderMinLength
+	}
+	// Sum over header without State and IntegrityCheck.File.
+	// To do that we just sum over the whole header and subtract.
+	var sum uint8
```
insomniacslk: I am not familiar with how the checksum should be computed here, so I trust what you're doing, but if you have a pointer to the specs about this I appreciate

insomniacslk: maybe split this into a `Checksum` method?

GanShun: Done. It's now in uefi.go

PR #25, opened 2018-06-28
Adding basic Assemble subcommand

GanShun:

### DIFF
--- a/uefi/flash.go
+++ b/uefi/flash.go
@@ -4,9 +4,11 @@ import (
 	"bytes"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
+	"sort"
 )
 
 // FlashSignature is the sequence of bytes that a Flash image is expected to
@@ -18,6 +20,8 @@ var (
 const (
 	// FlashDescriptorLength represents the size of the descriptor region.
 	FlashDescriptorLength = 0x1000
+	// FlashSignatureLength represents the size of the flash signature
+	FlashSignatureLength = 4
 )
 
 // FlashDescriptor is the main structure that represents an Intel Flash Descriptor.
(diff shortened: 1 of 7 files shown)


uefi/flash.go
```diff
@@ -35,6 +37,57 @@ type FlashDescriptor struct {
...
 }
 
+func findSignature(buf []byte) (int, error) {
+	if bytes.Equal(buf[16:16+len(FlashSignature)], FlashSignature) {
+		// 16 + 4 since the descriptor starts after the signature
+		return 20, nil
+	}
+	if bytes.Equal(buf[:len(FlashSignature)], FlashSignature) {
+		// + 4 since the descriptor starts after the signature
+		return 4, nil
```
rminnich: is it worth defining a nice named constant?

uefi/flash.go
```diff
@@ -35,6 +37,57 @@ type FlashDescriptor struct {
...
+		return err
+	}
+	fd.DescriptorMapStart = uint(descriptorMapStart)
+
+	// Descriptor Map
+	desc, err := NewFlashDescriptorMap(fd.buf[fd.DescriptorMapStart : fd.DescriptorMapStart+FlashDescriptorMapSize])
+	if err != nil {
+		return err
+	}
+	fd.DescriptorMap = *desc
```
rminnich: Why isn't DescriptorMap just a pointer too? just wondering. This deref/copy seems more work than needed.

uefi/flash.go
```diff
@@ -139,6 +201,85 @@ func (f *FlashImage) Extract(dirPath string) error {
...
+	// We need to sort them since a) we don't really know the order until we parse the block numbers
+	// and b) the order may have changed anyway.
+	if !f.IFD.Region.BIOS.Valid() {
+		return nil, fmt.Errorf("no BIOS region: invalid region parameters %v", f.IFD.Region.BIOS)
+	}
+	type region struct {
+		P   *Region
+		buf []byte
+	}
+	regions := make([]region, 0, 0)
```
rminnich: I can see the length as 0, but why not make the capacity something reasonable, or just not set it? Or just do
var regions []region
if you're going to make it zero size?
If you do this:
regions := make([]region,0, 32) 
then the append will be pretty efficient until you hit 32. No reallocations until you hit it.

uefi/gberegion.go
```diff
@@ -45,3 +46,14 @@ func (gbe *GBERegion) Extract(parentPath string) error {
 	gbe.ExtractPath, err = ExtractBinary(gbe.buf, dirPath, "gberegion.bin")
 	return err
 }
+
+// Assemble assembles the GBE Region from the binary file.
+func (gbe *GBERegion) Assemble() ([]byte, error) {
+	// We have to trust that the ExtractPath we read from the JSON is correct.
```
rminnich: I'm getting used to this comment everywhere :-)

uefi/flash.go
GanShun: ok will do

uefi/gberegion.go
GanShun: ok I'll remove it. :P

uefi/flash.go
GanShun: ok

uefi/flash.go
GanShun: I'll do it.

uefi/flash.go
GanShun: done

uefi/flash.go
GanShun: There are max 4 regions. I'll make cap 4.

PR #26, opened 2018-06-29
Add Assemble for Firmware Volumes

GanShun:
Also fix a bug where Extract was called on a copy instead of the actual
struct.

### DIFF
--- a/uefi/firmwarevolume.go
+++ b/uefi/firmwarevolume.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
+	"io/ioutil"
 	"path/filepath"
 
 	uuid "github.com/linuxboot/fiano/uuid"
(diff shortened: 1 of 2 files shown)


uefi/firmwarevolume.go
```diff
@@ -177,6 +178,20 @@ func (fv *FirmwareVolume) Validate() []error {
 	return errs
 }
 
+// Assemble assembles the Firmware Volume from the binary file.
+// TODO: HANDLE HEADER CHANGES.
+// We assume the FV length hasn't changed, and we assume the FV offset is the same as specified in
+// the JSON. We also don't check that the extended header or data offset is the same.
+// This is not something that's expected to change easily.
+func (fv *FirmwareVolume) Assemble() ([]byte, error) {
```
insomniacslk: this method's name is a bit misleading because there is no actual assembling operation in here, but I think we can go with this anyway as the intent is obvious and explained in the doc

PR #27, opened 2018-07-02
Add Assemble for FV Files

GanShun:

With this, we should be able to extract and reassemble winterfell down to the file level, haven't checked on winterfell yet though.

### DIFF
--- a/uefi/firmwarefile.go
+++ b/uefi/firmwarefile.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
+	"io/ioutil"
 
 	uuid "github.com/linuxboot/fiano/uuid"
 )
@@ -96,6 +97,16 @@ type FirmwareFile struct {
 
 	//Metadata for extraction and recovery
 	ExtractPath string
+}
+
+// Assemble assembles the Firmware File from the binary
+func (f *FirmwareFile) Assemble() ([]byte, error) {
+	var err error
+	f.buf, err = ioutil.ReadFile(f.ExtractPath)
+	if err != nil {
+		return nil, err
+	}
+	return f.buf, nil
 }
 
 // Extract extracts the FFS to the directory passed in.
--- a/uefi/firmwarevolume.go
+++ b/uefi/firmwarevolume.go
@@ -115,7 +115,7 @@ func (fv *FirmwareVolume) Extract(parentPath string) error {
 		return err
 	}
 
-	// Extract all filess.
+	// Extract all files.
 	for _, f := range fv.Files {
 		if err = f.Extract(dirPath); err != nil {
 			return err
@@ -178,6 +178,12 @@ func (fv *FirmwareVolume) Validate() []error {
 	return errs
 }
 
+func fillFFs(b []byte) {
+	for i := range b {
+		b[i] = 0xFF
+	}
+}
+
 // Assemble assembles the Firmware Volume from the binary file.
 // TODO: HANDLE HEADER CHANGES.
 // We assume the FV length hasn't changed, and we assume the FV offset is the same as specified in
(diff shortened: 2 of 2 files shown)


uefi/firmwarefile.go
```diff
@@ -98,6 +99,16 @@ type FirmwareFile struct {
 	ExtractPath string
 }
 
+// Assemble assembles the Firmware File from the binary
+func (f *FirmwareFile) Assemble() ([]byte, error) {
+	var err error
```
rjoleary: I think lines 104-109 could simply be:

    return ioutil.ReadFile(f.ExtractPath)

rjoleary: Never mind, I see that you're trying to store into `f.buf`.

uefi/firmwarevolume.go
```diff
@@ -189,6 +190,39 @@ func (fv *FirmwareVolume) Assemble() ([]byte, error) {
...
+	for _, f := range fv.Files {
+		fBuf, err := f.Assemble()
+		if err != nil {
+			return nil, err
+		}
+		fLen := uint64(len(fBuf))
+		if fLen+fOffset > fvLen {
+			// TODO: Actually loop through and calculate the full size so we know how much to reduce by.
+			// For now we just return early
+			return nil, errors.New("insufficient space in FV, files too big")
```
rjoleary: Can you include the FV size in the diagnosis message? Something like:

    insufficient space in 43.5KiB FV, files too big

uefi/firmwarevolume.go
```diff
@@ -189,6 +190,39 @@ func (fv *FirmwareVolume) Assemble() ([]byte, error) {
...
+	// Make sure we don't go over the size.
+	fvLen := uint64(len(fv.buf))
+	fOffset := fv.DataOffset
+	for _, f := range fv.Files {
+		fBuf, err := f.Assemble()
+		if err != nil {
+			return nil, err
+		}
+		fLen := uint64(len(fBuf))
+		if fLen+fOffset > fvLen {
```
rjoleary: Should this check be moved until after the file is aligned because it might be possible that aligning the file might overflow the FV?

uefi/firmwarevolume.go
```diff
@@ -189,6 +190,39 @@ func (fv *FirmwareVolume) Assemble() ([]byte, error) {
...
+		// We have to pad to the 8 byte alignments.
+		for alignBytes := align8(fOffset) - fOffset; alignBytes > 0; alignBytes-- {
+			fv.buf[fOffset] = 0xFF
+			fOffset++
+		}
+		copy(fv.buf[fOffset:], fBuf)
+		fOffset += fLen
+	}
+
+	if fOffset == fvLen {
```
rjoleary: This if-statement doesn't look necessary because the following for-loop is skipped and the same values are returned when it is false.

uefi/firmwarevolume.go
GanShun: Nice catch! I'll change it

uefi/firmwarevolume.go
GanShun: done

uefi/firmwarevolume.go
```diff
@@ -189,6 +190,39 @@ func (fv *FirmwareVolume) Assemble() ([]byte, error) {
...
+		// We're done here, no need to fill the rest of the fv
+		return fv.buf, nil
+	}
+
+	// Fill to the end with FFs
+	for ; fOffset < fvLen; fOffset++ {
+		fv.buf[fOffset] = 0xFF
```
rjoleary: Since you're filling with 0xFF in multiple places, you can add a function:

```
func fillFF(data []byte) {
	for i := range data {
		data[i] = 0xff
	}
}
```

And call like this:

```
fillFF(fv.buf[fOffset:fvLen])
```

And on line 209, you can use:

```
fillFF(fv.buf[fOffset:align8(fOffset)])
fOffset = align8(fOffset)
```

uefi/firmwarevolume.go
GanShun: oo ok thanks!

uefi/firmwarevolume.go
GanShun: I could, but I just realized that for the alignments, I have to check if the filling of FFs will go out of bounds. I'll redo it slightly differently

GanShun: done

uefi/firmwarevolume.go
```diff
@@ -189,6 +195,36 @@ func (fv *FirmwareVolume) Assemble() ([]byte, error) {
...
+			// For now we just return early
+			return nil, fmt.Errorf("insufficient space in %#x bytes FV, files too big", fvLen)
+		}
+		// Overwrite old data in the firmware volume.
+		copy(fv.buf[fOffset:], fBuf)
+		fOffset += fLen
+	}
+
+	// Fill to the end with FFs
+	fillFFs(fv.buf[fOffset:fvLen])
```
rjoleary: I'm sorry, because of the last refactor, you actually need an if-statement here now:

```
if fOffset < fvLen {
    fillFFs(fv.buf[fOffset:fvLen])
}
```

GanShun: Done

PR #33, opened 2018-07-09
Integration tests

rjoleary:
Feel free to add more ROMs to `integration/roms`. The test uses all ROMs in that directory which end in `.rom`. ROMs which begin with `SECRET` are gitignore'd.

Right now, the test fails because there is a difference:

```
Binary files /tmp/utk-test631886125/dir1/bios/0x0/fv.bin and /tmp/utk-test631886125/dir2/bios/0x0/fv.bin differ
Binary files /tmp/utk-test631886125/dir1/bios/biosregion.bin and /tmp/utk-test631886125/dir2/bios/biosregion.bin differ
```

I will look more into why the difference exists later this week with vbindiff.

### DIFF
--- a/integration/utk_test.go
+++ b/integration/utk_test.go
@@ -1,0 +1,134 @@
+package main
+
+import (
+	"encoding/json"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+)
+
+// Returns all the ROM names (files which end in .rom) inside the roms folder.
+func romList(t *testing.T) []string {
+	roms, err := filepath.Glob("roms/*.rom")
+	if err != nil {
+		t.Fatalf("could not glob roms/*.rom, %v", err)
+	}
+	if len(roms) == 0 {
+		t.Fatal("no ROMs found with roms/*.rom")
+	}
+	return roms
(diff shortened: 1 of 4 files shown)


rjoleary:
Travis is giving the error "Unknown firmware type". The two other PRs need to be merged first.

GanShun:
This should work now, please rebase on master and try again

integration/utk_test.go
```diff
@@ -0,0 +1,134 @@
...
+//
+// 1. utk extract tt.rom dir1
+// 2. utk assemble dir1 tmp.rom
+// 3. utk extract tmp.rom dir2
+//
+// The test passes iff the contents or dir1 and dir2 recursively equal. This
+// roundabout method is used because UTK can re-assemble a ROM image which is
+// logically equal to the original, but not bitwise equal (due to a different
+// compression algorithm being used). To compare the ROMs logically, step 3 is
+// required to decompresses it.
```
GanShun: We'll have to fix some stuff to make this actually work for compression. For example, I dump the fv binary before decompression. So technically the binary files extracted in step 1 has some files still compressed. We need to ignore those in the future.

PR #40, opened 2018-07-18
Parse only sections of files we support

GanShun:

### DIFF
--- a/uefi/firmwarefile.go
+++ b/uefi/firmwarefile.go
@@ -40,6 +40,24 @@ const (
 	fvFileTypeFFSMax   FVFileType = 0xFF
 )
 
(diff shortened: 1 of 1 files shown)


uefi/firmwarefile.go
```diff
@@ -40,6 +40,27 @@ const (
 	fvFileTypeFFSMax   FVFileType = 0xFF
 )
 
+var supportedFiles map[FVFileType]bool
+
+func init() {
+	// These are the file types that we'll actually try to parse sections for.
```
insomniacslk: as discussed in chat, make this a statically defined map

GanShun: done

PR #42, opened 2018-07-20
File and Section Assembly

GanShun:

### DIFF
--- a/uefi/filesection.go
+++ b/uefi/filesection.go
@@ -37,9 +37,10 @@ type FileSection struct {
 
 	//Metadata for extraction and recovery
 	ExtractPath string
+	fileOrder   int
 }
 
-// Assemble assembles the Firmware File from the binary
+// Assemble assembles the section from the binary
 func (f *FileSection) Assemble() ([]byte, error) {
 	var err error
 	f.buf, err = ioutil.ReadFile(f.ExtractPath)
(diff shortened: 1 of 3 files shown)


uefi/firmwarefile.go
```diff
@@ -115,13 +139,95 @@ type FirmwareFile struct {
...
+		f.buf, err = ioutil.ReadFile(f.ExtractPath)
+		if err != nil {
+			return nil, err
+		}
+		return f.buf, nil
+	}
+
+	// Otherwise, we reconstruct the entire file from the sections and the
+	// file header using data from the JSON. This means that some JSON values
+	// will now be respected, including GUID changes. However file lengths and
```
rjoleary: Make this present tense: This means that some JSON values _are_ respected ...

uefi/firmwarefile.go
```diff
@@ -115,13 +139,95 @@ type FirmwareFile struct {
...
+	// checksums will be recalculated.
+
+	// Assemble all sections so we know the final file size. We need to do this
+	// to know if we need to use the extended header.
+	fileData := []byte{}
+	dLen := uint64(0)
```
rjoleary: What is the purpose of `dLen`? Won't it always equal `len(fileData)` after any operation?

uefi/filesection.go
```diff
@@ -37,9 +37,10 @@ type FileSection struct {
 
 	//Metadata for extraction and recovery
 	ExtractPath string
+	fileOrder   int
 }
 
-// Assemble assembles the Firmware File from the binary
+// Assemble assembles the section from the binary
```
rjoleary: Why doesn't assemble read the `fileOrder`? How does it get the sections back into the original order?

rjoleary: Never mind, I see that the information is stored in the JSON.

PR #48, opened 2018-07-25
fmap: Import from u-root

rjoleary:


rjoleary:
u-root: https://github.com/u-root/u-root/pull/839

insomniacslk:
Looks good to me. Just to understand, what's the plan to keep this in sync with u-root's fmap?

rjoleary:
@insomniacslk The plan is to also remove fmap from u-root, so it is only contained in fiano.

See: https://github.com/u-root/u-root/pull/839

PR #65, opened 2018-08-14
LZMA

rjoleary:


pkg/lzma/lzma.go
```diff
@@ -0,0 +1,46 @@
...
+func Decode(encodedData []byte) ([]byte, error) {
+	cmd := exec.Command(lzmaDecode)
+	cmd.Stdin = bytes.NewBuffer(encodedData)
+	cmd.Stderr = os.Stderr
+	return cmd.Output()
+}
+
+// Encode encodes a byte slice with LZMA.
+func Encode(decodedData []byte) ([]byte, error) {
+	return EncodeLevel(decodedData, 6)
```
GanShun: why default to 6? is that the default for the regular lzma command? I thought it was 7

uefi/section.go
```diff
@@ -71,18 +84,32 @@ type SectionExtHeader struct {
...
 	fileOrder   int
 
 	// Type specific fields
+	TypeSpecific interface{} `json:",omitempty"`
+
+	// For EFI_SECTION_USER_INTERFACE
 	Name string `json:",omitempty"`
+
+	// Encapsulated section (only for EFI_SECTION_COMPRESSION, EFI_SECTION_GUID_DEFINED and EFI_SECTION_DISPOSABLE)
+	Encapsulated *Section `json:",omitempty"`
```
GanShun: Are you sure it's always going to be a Section? I think it can be other things. you should probably use []Firmware

You also don't want to assume there's only one thing.

uuid/uuid.go
```diff
@@ -58,6 +59,15 @@ func Parse(s string) (*UUID, error) {
 	return &u, nil
 }
 
+// Parse parses a uuid string or panics.
```
GanShun: *MustParse

I'm surprised Lint didn't complain

pkg/lzma/lzma.go
rjoleary: This man pages says the default is -6: https://linux.die.net/man/1/lzma

uefi/section.go
```diff
@@ -71,18 +84,32 @@ type SectionExtHeader struct {
...
 	fileOrder   int
 
 	// Type specific fields
+	TypeSpecific interface{} `json:",omitempty"`
+
+	// For EFI_SECTION_USER_INTERFACE
 	Name string `json:",omitempty"`
+
+	// Encapsulated firmware
+	Encapsulated Firmware `json:",omitempty"`
```
GanShun: Actually I'm pretty sure I'm looking at something in one of my roms, where there's an lzmacustomdecompress section, with 3 things inside it.

pkg/lzma/lzma.go
GanShun: I see

uefi/section.go
GanShun: In the case of lzma decompression, we should be able to assume that these []Firmware elements are of type Section, but lets keep it as []Firmware so we can reuse the same field for the Volume Image sections as those contain FVs.

uefi/section.go
```diff
@@ -169,8 +196,39 @@ func NewSection(buf []byte, fileOrder int) (*Section, error) {
...
+		if err := binary.Read(r, binary.LittleEndian, typeSpec); err != nil {
+			return nil, err
+		}
+		s.TypeSpecific = &typeSpec
+
+		// Determine whether the encapsulated section needs processing.
+		if typeSpec.Attributes&uint16(GUIDEDSectionProcessingRequired) == 0 {
+			// No processing required
+			var err error
+			s.Encapsulated, err = NewSection(buf[typeSpec.DataOffset:], 0)
```
GanShun: If I'm right, and there can be multiple sections, this will need to mirror the looping construction behavior in File.

uefi/section.go
GanShun: Generalizing this for []Firmware for all sections also makes the visitor pattern application much nicer I think

rjoleary: I didn't know about lzmacustomdecompress. I'll make it an array.

pkg/lzma/lzma.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+)
+
+const (
+	lzmaEncode = "lzma"
```
hugelgupf: make these passable by flag?

hugelgupf: Maybe like this:

```go
// LZMA configures the LZMA binary to use.
type LZMA struct {
  Encoder, Decoder string
}

// Use whatever is in PATH by default.
var Default = LZMA{
  Encoder: "lzma",
  Decoder: "unlzma",
}

func (l LZMA) Decode(...) {
  cmd := exec.Command(l.Decoder) ...
}

// easy default func
func Decode(...) {
  return Default.Decode(...)
}

func NewLZMA(encoder, decoder string) LZMA {
  
}
```

hugelgupf: Then, a command can pass the flag to NewLZMA, or use Default by default.

uefi/section.go
rjoleary: Done!

rjoleary: We agreed to do this in another PR.

pkg/lzma/lzma.go
rjoleary: Thanks, I like this idea! Will do in another PR.

rjoleary: #68

PR #66, opened 2018-08-14
Visitor pattern prototype

rjoleary:

### DIFF
--- a/uefi/firmwarevolume.go
+++ b/uefi/firmwarevolume.go
@@ -98,6 +98,21 @@ type FirmwareVolume struct {
 	ExtractPath string
 }
 
+// Apply calls the visitor on the FirmwareVolume.
+func (fv *FirmwareVolume) Apply(v Visitor) error {
+	return v.VisitFV(fv)
+}
+
+// ApplyChildren calls the visitor on each child node of FirmwareVolume.
+func (fv *FirmwareVolume) ApplyChildren(v Visitor) error {
+	for _, f := range fv.Files {
+		if err := v.VisitFile(f); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 // GetErasePolarity gets the erase polarity
 func (fv *FirmwareVolume) GetErasePolarity() uint8 {
 	if fv.Attributes&0x800 != 0 {
--- a/uefi/flash.go
+++ b/uefi/flash.go
@@ -52,6 +52,16 @@ func FindSignature(buf []byte) (int, error) {
 		hex.Dump(buf[:20]))
 }
 
+// Apply calls the visitor on the FlashDescriptor.
+func (fd *FlashDescriptor) Apply(v Visitor) error {
+	return v.VisitIFD(fd)
+}
+
+// ApplyChildren calls the visitor on each child node of FlashDescriptor.
+func (fd *FlashDescriptor) ApplyChildren(v Visitor) error {
+	return nil
+}
+
 // ParseFlashDescriptor parses the ifd from the buffer
 func (fd *FlashDescriptor) ParseFlashDescriptor() error {
 	if buflen := len(fd.buf); buflen != FlashDescriptorLength {
@@ -139,6 +149,36 @@ type FlashImage struct {
 	// Metadata for extraction and recovery
 	ExtractPath string
 	regions     []Firmware
+}
+
+// Apply calls the visitor on the FlashImage.
+func (f *FlashImage) Apply(v Visitor) error {
+	return v.VisitImage(f)
+}
+
+// ApplyChildren calls the visitor on each child node of FlashImage.
+func (f *FlashImage) ApplyChildren(v Visitor) error {
+	if f.BIOS != nil {
+		if err := v.VisitBIOSRegion(f.BIOS); err != nil {
+			return err
+		}
+	}
+	if f.ME != nil {
+		if err := v.VisitMERegion(f.ME); err != nil {
+			return err
+		}
+	}
+	if f.GBE != nil {
+		if err := v.VisitGBERegion(f.GBE); err != nil {
+			return err
+		}
+	}
+	if f.PD != nil {
+		if err := v.VisitPDRegion(f.PD); err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 // IsPCH returns whether the flash image has the more recent PCH format, or not.
--- a/uefi/visitor.go
+++ b/uefi/visitor.go
@@ -1,0 +1,28 @@
+package uefi
+
+// Visitor represents an operation which can be applied to the Firmware. To
+// implement a visitor, implement a method for each concrete Firmware type you
+// want to visit. The remaining nodes can be recursed over automatically with:
+//
+//     func (v *Find) VisitFV(fv *uefi.FirmwareVolume) error {
+//             return fv.ApplyChildren(v)
+//     }
+//
+// Or pruned with:
+//
+//     func (v *Find) VisitFV(fv *uefi.FirmwareVolume) error {
+//             return nil
+//     }
+type Visitor interface {
+	VisitImage(*FlashImage) error
+	VisitFV(*FirmwareVolume) error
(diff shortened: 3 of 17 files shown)


visitors/README.md
```diff
@@ -0,0 +1,24 @@
...
+      save winterfell2.rom
+
+Operations:
+
+- `remove (GUID|NAME)`: Remove the first file which matches the given GUID or
+  NAME. If the number of matches is unequal to one, an error results.
+- `replace (GUILD|NAME) FILE`: Replace the first file which matches the given
+  GUID or NAME with the contents of FILE. If the number of matches is unequal
+  to one, an error results.
+- `assemble FILE`: Re-assemble the BIOS and save to the given file.
```
GanShun: assemble should have some directory to assemble from? also it doesn't make sense to pass utk a bios image if you already want to assemble it from the working directory

GanShun:
Overall I really like this design, it's great, just some nits

uefi/ast.go
```diff
@@ -0,0 +1,6 @@
+package uefi
+
+type ASTNode interface {
+	Apply(v Visitor) error
+	ApplyChildren(v Visitor) error
```
hugelgupf: ApplyRecursive? Dunno, I don't feel strongly.

uefi/ast.go
```diff
@@ -0,0 +1,6 @@
+package uefi
+
+type ASTNode interface {
```
hugelgupf: // ASTNode is a ...

uefi/visitor.go
```diff
@@ -0,0 +1,8 @@
+package uefi
+
+type Visitor interface {
```
hugelgupf: Can we put this in ast.go as well? the ASTNode interface makes little sense without the Visitor interface right there.

hugelgupf: // Visitor is...

uefi/flash.go
```diff
@@ -141,6 +141,21 @@ type FlashImage struct {
 	regions     []Firmware
 }
 
+func (n *FlashImage) Apply(v Visitor) error {
+	return v.VisitImage(n)
+}
+
+func (n *FlashImage) ApplyChildren(v Visitor) error {
+	if n.BIOS != nil {
```
hugelgupf: Why are we only applying this to BIOSRegion? Shouldn't we visit all children? IFD, BIOS, ME, GBE, PD?

uefi/firmwarevolume.go
```diff
@@ -113,6 +113,19 @@ type FirmwareVolume struct {
 	ExtractPath string
 }
 
+func (n *FirmwareVolume) Apply(v Visitor) error {
+	return v.VisitFV(n)
+}
+
+func (n *FirmwareVolume) ApplyChildren(v Visitor) error {
+	for _, f := range n.Files {
```
hugelgupf: why only visit files, not blocks?

visitors/cli.go
```diff
@@ -0,0 +1,56 @@
...
+	visitorRegistry[name] = visitorEntry{
+		numArgs:       numArgs,
+		createVisitor: createVisitor,
+	}
+}
+
+// TODO: display some type of help message
+func ParseCLI(args []string) ([]uefi.Visitor, error) {
+	visitors := []uefi.Visitor{}
+	for i := 0; i < len(args); i++ {
```
hugelgupf: Instead of just using i, can we slide across args for readability? E.g.

```go
for len(args) > 0 {
  o, ok := visitorRegistry[args[0]]
  ..
  if o.numArgs < len(args) {
    too few args
  }
  cmdArgs := args[0:o.numArgs]
  ...
  args = args[o.numArgs:]
}
```

this won't affect the caller, since we are only modifying the slice header.

uefi/firmwarevolume.go
```diff
@@ -113,6 +113,19 @@ type FirmwareVolume struct {
 	ExtractPath string
 }
 
+func (n *FirmwareVolume) Apply(v Visitor) error {
+	return v.VisitFV(n)
+}
+
+func (n *FirmwareVolume) ApplyChildren(v Visitor) error {
```
hugelgupf: Should ApplyChildren also apply to self? Depending on what you choose, document it in the interface docs for ASTNode.

hugelgupf: I also get the feeling that ApplyChildren could just be a Visitor implementation. E.g.

```go
// RecursiveVisitor is a visitor that recursively applies the nested Visitor to all
// children nodes of each visited node and the visited node itself.
type RecursiveVisitor struct {
  nested Visitor
}

func (rv RecursiveVisitor) VisitFV(fv *FirmwareVolume) error {
  rv.nested.VisitFV(fv)

  for _, f := range fv.Files {
    rv.VisitFile(f)
  }
}

...
```

uefi/flash.go
```diff
@@ -141,6 +141,21 @@ type FlashImage struct {
...
+		for _, fv := range n.BIOS.FirmwareVolumes {
```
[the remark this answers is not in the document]
rjoleary: Done!

utk2/utk2.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	// TODO: This is temporary code to turn `parsedRoot` into an ASTNode
+	// interface. Ideally, `uefi.Parse` will return an ASTNode directly.
+	var astRoot uefi.ASTNode
+	switch x := parsedRoot.(type) {
+	case *uefi.FlashImage:
+		astRoot = x
+	case *uefi.BIOSRegion:
+		astRoot = &uefi.FlashImage{BIOS: x}
```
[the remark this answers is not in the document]
rjoleary: Done!

visitors/README.md
rjoleary: I renamed this to `save`. Read the new README and see if that makes sense.

uefi/ast.go
rjoleary: It only applies to the children, so I feel the name is apt.

uefi/ast.go
rjoleary: Done!

uefi/visitor.go
rjoleary: I deleted `ASTNode` altogether.

rjoleary: Done!

uefi/flash.go
rjoleary: Done! Although, the visitors a bit verbose now. I might make a separate Visitor interface for just the intel nodes.

visitors/cli.go
rjoleary: Thanks, I'll do this!

uefi/firmwarevolume.go
rjoleary: Documented! If I made it call itself, a lot of its uses would turn into infinite recursion. `ApplyChildren` is sort of the default function to call when a visitor skips over a node.

visitors/cli.go
rjoleary: Done!

uefi/visitor.go
```diff
@@ -0,0 +1,28 @@
...
+
+// Visitor represents an operation which can be applied to the Firmware. To
+// implement a visitor, implement a method for each concrete Firmware type you
+// want to visit. The remaining nodes can be recursed over automatically with:
+//
+//     func (v *Find) VisitFV(fv *uefi.FirmwareVolume) error {
+//             return fv.ApplyChildren(v)
+//     }
+//
+// Or pruned with:
```
hugelgupf: I don't see the logic for pruning? 

In any case, I think pruning should rather happen with something like `uefi.ErrRemove`. People might return `nil` for "do nothing" or "I successfully collected some metadata from this item".

rjoleary: Returning nil will not modify the tree at all. Returning `nil` does nothing which includes not recusing over children.

Maybe pruning is the wrong word. I got it from the CS idea of "pruning a search space".

PR #68, opened 2018-08-15
lzma: Use non-streamed mode

rjoleary:

### DIFF
--- a/pkg/lzma/lzma.go
+++ b/pkg/lzma/lzma.go
@@ -13,33 +13,39 @@ import (
 	"os/exec"
 )
 
-const (
-	lzmaEncode = "lzma"
-	lzmaDecode = "unlzma"
-)
+// LZMA configures the LZMA binary to use.
+type LZMA struct {
+	Encoder, Decoder string
+}
+
+// Default uses whichever commands are in $PATH.
+var Default = LZMA{
+	Encoder: "lzma",
+	Decoder: "unlzma",
+}
 
 // Decode decodes a byte slice of LZMA data.
-func Decode(encodedData []byte) ([]byte, error) {
-	cmd := exec.Command(lzmaDecode)
+func (l LZMA) Decode(encodedData []byte) ([]byte, error) {
+	cmd := exec.Command(l.Decoder)
 	cmd.Stdin = bytes.NewBuffer(encodedData)
 	cmd.Stderr = os.Stderr
 	return cmd.Output()
 }
 
 // Encode encodes a byte slice with LZMA.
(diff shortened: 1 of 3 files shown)


rjoleary:
Travis's "file" command is different from my "file" command.

pkg/lzma/lzma.go
```diff
@@ -7,40 +7,97 @@ package lzma
...
+// backwards compatible with lzma (assume the --format=lzma flag is used). And,
+// on many distros, lzma points to xz. XZ is the new LZMA.
+var Default = LZMA{
+	XZPath: "xz",
+}
 
 // Decode decodes a byte slice of LZMA data.
-func Decode(encodedData []byte) ([]byte, error) {
-	cmd := exec.Command(lzmaDecode)
+func (l LZMA) Decode(encodedData []byte) ([]byte, error) {
```
rminnich: l LZMA? Why not a pointer?

pkg/lzma/lzma.go
```diff
@@ -7,40 +7,97 @@ package lzma
...
-	return cmd.Output()
+	out, err := cmd.Output()
+	if err == nil {
+		return out, nil
+	}
+
+	// TODO: This is extremely hacky. This reverses the damage we inflicted
+	// on the compressed lzma file in EncodeLevel.
+	copy(encodedData[5:5+8], []byte{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff})
+	cmd = exec.Command(l.XZPath, args...)
```
hugelgupf: Why xz decompress twice? If there's a good reason (I don't really get it), please add a comment explaining this.

rjoleary: I think it is a bug in xz. I'm going to look into it more today.

hugelgupf: wat

rjoleary:
Abandoning in favour of #90

pkg/lzma/lzma.go
rjoleary: Better explanation: When the xz command compresses, it stores the size at the end of the file and 0xFFFFFFFFFFFFFFFF in the header. EDK2's decompressor is primitive and will try to allocate 0xFFFFFFFFFFFFFFFF bytes and fail. So, the Encode function writes the size in the header which works for EDK2's tool. Unfortunately, xz considers an lzma file which has a size in both places corrupt, but will still decompress it and return exit status 1 (false positive). The second decompression overwrites the header to see if first's exit status is a false positive.

We're not going to do this and instead use the Go implementation of LZMA and upstream any issues we encounter.

PR #70, opened 2018-08-15
Fix issues from the go report card

rjoleary:


utk/utk.go
```diff
@@ -218,6 +218,9 @@ func (a *assembleCmd) Execute(_ context.Context, f *flag.FlagSet, _ ...interface
 		return subcommands.ExitFailure
 	}
 	jsonbuf, err := ioutil.ReadFile("summary.json")
+	if err != nil {
+		log.Fatal(err)
```
GanShun: do a log.Print and a subcommands.ExitFailure instead

rjoleary: Address is #76

PR #89, opened 2018-08-21
visitor: Simplify the interface with type switch

rjoleary:
Most of the visitors only needed to implement a handful of the
functions. The remaining functions have some default behaviour.

This change simplifies creating new visitors because the default
behaviour can be placed in the default case.

### DIFF
--- a/visitors/table.go
+++ b/visitors/table.go
@@ -19,6 +19,33 @@ func indent(n int) string {
 	return strings.Repeat(" ", n)
 }
 
+// Visit applies the Table visitor to any Firmware type.
+func (v *Table) Visit(f uefi.Firmware) error {
+	switch f := f.(type) {
+	case *uefi.FlashImage:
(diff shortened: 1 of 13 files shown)


visitors/table.go
```diff
@@ -19,6 +19,33 @@ func indent(n int) string {
...
+	case *uefi.BIOSRegion:
+		return v.printRow(f, "BIOS", "", "", "")
+	case *uefi.MERegion:
+		return v.printRow(f, "ME", "", "", "")
+	case *uefi.GBERegion:
+		return v.printRow(f, "GBE", "", "", "")
+	case *uefi.PDRegion:
+		return v.printRow(f, "PD", "", "", "")
+	default:
+		return v.printRow(f, "Err", "", "", "")
```
GanShun: one thing you could do here is instead, print unknown firmware type, and use reflection to display the type. That'll be more useful than just an Err

PR #91, opened 2018-08-22
Add a "remove" visitor

rjoleary:


visitors/remove.go
```diff
@@ -0,0 +1,40 @@
...
+// Remove all firmware files with the given GUID.
+type Remove struct {
+	// Input
+	Predicate func(f *uefi.File) bool
+
+	// Output
+	Matches []*uefi.File
+}
+
+// Visit applies the Remove visitor to any Firmware type.
```
GanShun: Can we actually try to remove by name as well? is it possible to use the Find visitor here?

rjoleary: Sounds good! I have to implement the before/after first.

PR #96, opened 2018-08-23
LZMA x86

rjoleary:


uefi/section.go
```diff
@@ -356,19 +361,21 @@ func NewSection(buf []byte, fileOrder int) (*Section, error) {
...
-				} else {
-					typeSpec.Compression = "LZMA"
-				}
+			case lzmaX86GUID:
+				typeSpec.Compression = "LZMAX86"
+				encapBuf, err = lzma.DecodeX86(buf[typeSpec.DataOffset:])
 			default:
 				typeSpec.Compression = "UNKNOWN"
 			}
+			if err != nil {
```
GanShun: I feel like if there's an error decoding, you should instead return the error rather than hide it

rjoleary: Instead, can we print the error as a warning and continue? The behaviour in UEFITool is to print a warning message when it cannot be decompressed and mark it as an unknown compression algorithm. This is nice because there is a variety of compression algos out there and we would still be able to partially extract to image.

GanShun: ok fine

PR #118, opened 2018-08-30
Assemble Visitor

GanShun:

### DIFF
--- a/visitors/assemble.go
+++ b/visitors/assemble.go
@@ -1,0 +1,349 @@
+// Copyright 2018 the LinuxBoot Authors. All rights reserved
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package visitors
+
+import (
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"log"
+	"sort"
+
+	"github.com/linuxboot/fiano/pkg/lzma"
+	"github.com/linuxboot/fiano/uefi"
+)
+
+// Assemble reconstitutes the firmware tree assuming that the leaf node buffers are accurate
+type Assemble struct {
+}
+
+// Run just applies the visitor.
+func (v *Assemble) Run(f uefi.Firmware) error {
+	return f.Apply(v)
+}
+
+// Visit applies the Assemble visitor to any Firmware type.
+func (v *Assemble) Visit(f uefi.Firmware) error {
+	var err error
+
+	// Get the damn Erase Polarity
+	if f, ok := f.(*uefi.FirmwareVolume); ok {
+		// Set Erase Polarity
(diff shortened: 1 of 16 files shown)


visitors/assemble.go
```diff
@@ -0,0 +1,359 @@
...
+func (v *Assemble) Run(f uefi.Firmware) error {
+	return f.Apply(v)
+}
+
+// Visit applies the Assemble visitor to any Firmware type.
+func (v *Assemble) Visit(f uefi.Firmware) error {
+	var err error
+
+	// Get the damn Erase Polarity
+	switch f := f.(type) {
```
rjoleary: Use `if ok, f := f.(*uefi.FirmwareVolume); ok {`

visitors/assemble.go
```diff
@@ -0,0 +1,359 @@
...
+
+	case *uefi.FirmwareVolume:
+		// Set Erase Polarity
+		uefi.Attributes.ErasePolarity = f.GetErasePolarity()
```
rjoleary: Rather than using a global variable, is it possible to store the erase polarity inside of the visitor instead?

visitors/assemble.go
```diff
@@ -0,0 +1,359 @@
...
+			}
+		default:
+			f.SetBuf(secData)
+		}
+
+		// Fix up the header
+		err = f.GenSecHeader()
+
+	case *uefi.FlashDescriptor:
+		if err = f.ParseFlashDescriptor(); err != nil {
```
rjoleary: You can just do `err = f.ParseFlashDescriptor()` because you return `err` at the end.

visitors/assemble.go
GanShun: I could store it in the visitor, but I think I call into code in package uefi that checks that variable so I would have to set it anyway. We should overhaul this, but not in this PR

visitors/assemble.go
GanShun: Done

visitors/save.go
```diff
@@ -0,0 +1,37 @@
+
```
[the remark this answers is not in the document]
GanShun: Done

visitors/assemble.go
```diff
@@ -0,0 +1,359 @@
...
+			// TODO: handle different sizes.
+			// We'll have to FF out the new regions/ check for clashes
+			copy(fBuf[fv.FVOffset:fv.FVOffset+fv.Length], fv.Buf())
+		}
+		// Set the buffer
+		f.SetBuf(fBuf)
+
+		return nil
+
+	case *uefi.GBERegion:
```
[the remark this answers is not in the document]
GanShun: Done

GanShun:
Yes those 3 are the goals for this PR

GanShun:
Will do

PR #136, opened 2018-08-31
Handle BIOS padding

GanShun:


pkg/visitors/extract.go
```diff
@@ -127,6 +127,10 @@ func (v *Extract) Visit(f uefi.Firmware) error {
 	case *uefi.PDRegion:
 		v2.DirPath = filepath.Join(v.DirPath, "pd")
 		f.ExtractPath, err = uefi.ExtractBinary(f.Buf(), v2.DirPath, "pdregion.bin")
+
+	case *uefi.BIOSPadding:
```
rjoleary: Also add entry to the table visitor.

PR #138, opened 2018-09-04
travis: stop testing against go tip

rminnich:
It's not a good idea, go tip is always in motion and you get
false negatives.


rminnich:
if you want to add 1.11 at some point, that's fine, but not go tip.

GanShun:
We're going to remove travis soon since we're moving to circleci, but this is ok for now.

PR #141, opened 2018-09-04
add a cat command to visitor package

rminnich:
This is a lot like replace. It will cat all files which match a GUID to a byte slice.

### DIFF
--- a/pkg/visitors/cat.go
+++ b/pkg/visitors/cat.go
@@ -1,0 +1,83 @@
+// Copyright 2018 the LinuxBoot Authors. All rights reserved
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package visitors
+
+import (
+	"io"
+	"os"
+	"regexp"
+
+	"github.com/linuxboot/fiano/pkg/uefi"
+)
+
+// Cat concatenates all RAW data sections from a file into a single byte slice.
+type Cat struct {
+	// Input
+	Predicate func(f *uefi.File, name string) bool
+
+	// Output
+	io.Writer
+	Matches []*uefi.File
+}
+
+// Run wraps Visit and performs some setup and teardown tasks.
+func (v *Cat) Run(f uefi.Firmware) error {
+	// First run "find" to generate a list of matches to replace.
+	find := Find{
+		Predicate: v.Predicate,
+	}
+	if err := find.Run(f); err != nil {
+		return err
+	}
+
+	v.Matches = find.Matches
+	for _, m := range v.Matches {
+		if err := m.Apply(v); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Visit applies the Extract visitor to any Firmware type.
+func (v *Cat) Visit(f uefi.Firmware) error {
+	switch f := f.(type) {
+
+	case *uefi.File:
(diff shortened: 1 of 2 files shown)


rminnich:
I already know you'll make me put the data into a file. But I had to at least try.

pkg/visitors/cat.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/linuxboot/fiano/pkg/uefi"
+)
+
+// Cat concatenates all RAW data sections from a file into a single byte slice.
+type Cat struct {
+	// Input
+	Predicate func(f *uefi.File, name string) bool
+
+	// Output
+	Buf     []byte
```
rjoleary: Everything looks good, except this buffer never gets printed. Here's one way to fix it:

- Change this `Buf` to an `io.Writer`
- In the test initialize it to a `bytes.Buffer`
- In the RegisterCLI, initialize it to `os.stdout`

pkg/visitors/cat.go
```diff
@@ -0,0 +1,75 @@
...
+// Visit applies the Extract visitor to any Firmware type.
+func (v *Cat) Visit(f uefi.Firmware) error {
+	switch f := f.(type) {
+
+	case *uefi.File:
+		return f.ApplyChildren(v)
+
+	case *uefi.Section:
+		if f.Header.Type == uefi.SectionTypeRaw {
+			v.Buf = append(v.Buf, f.Buf()[4:]...)
```
rjoleary: Put a big TODO here about the header size being wrong.

pkg/visitors/cat.go
```diff
@@ -0,0 +1,75 @@
...
+		return f.ApplyChildren(v)
+
+	default:
+		// Must be applied to a File to have any effect.
+		return nil
+	}
+}
+
+func init() {
+	RegisterCLI("cat", 2, func(args []string) (uefi.Visitor, error) {
```
rjoleary: The number `2` on this line is the number of arguments. It should probably be 1.

pkg/visitors/cat_test.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	// Find the section and make sure it contains the expected data.
+	results := find(t, f, testGUID)
+	if len(results) != 1 {
+		t.Fatalf("got %d matches; expected 1", len(results))
+	}
+	got := results[0].Sections[0].Buf()
+	if !reflect.DeepEqual(got, catdat) {
+		t.Fatalf("want %v; got %v", catdat, got)
```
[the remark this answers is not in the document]
rminnich: done haha

pkg/visitors/cat.go
rminnich: My intent here is that some program will call this and manage the Buf by itself. I guess that does not with with how the commands work ... ok, I'll change it.

pkg/visitors/cat.go
rminnich: sure

pkg/visitors/cat.go
rminnich: gotcha.

pkg/visitors/cat.go
rminnich: done

rminnich:
Note the big test file is gone

PR #155, opened 2018-09-08
Count visitor

rjoleary:
Fixes https://github.com/linuxboot/fiano/issues/135


pkg/visitors/count_test.go
```diff
@@ -0,0 +1,41 @@
...
+// license that can be found in the LICENSE file.
+
+package visitors
+
+import (
+	"testing"
+)
+
+func TestCount(t *testing.T) {
+	f := parseImage(t)
```
GanShun: maybe we should move parseImage into another file rather than just keeping it in find_tests. maybe some file for visitor test helpers

rjoleary: See #158

PR #156, opened 2018-09-13
utk: add a help entry in the visitor struct, ListCLI,  and a help command

rminnich:

### DIFF
--- a/cmds/utk/utk.go
+++ b/cmds/utk/utk.go
@@ -55,6 +55,7 @@ package main
 
 import (
 	"flag"
+	"fmt"
 	"io/ioutil"
 	"log"
 	"os"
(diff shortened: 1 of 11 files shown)


cmds/utk/utk.go
```diff
@@ -69,6 +71,12 @@ func main() {
 		log.Fatal("at least one argument is required")
 	}
 
+	// help can't be registered as a visitor because it has nothing to visit.
+	if flag.NArg() == 1 && flag.Args()[0] == "help" {
```
rjoleary: Instead, can you put this in the `flag.Usage`? Like:

```
flag.Usage = func() {
   ...
}
```

rjoleary:
Thanks! This fixes #148

pkg/visitors/extract.go
```diff
@@ -141,7 +141,7 @@ func (v *Extract) Visit(f uefi.Firmware) error {
 
 func init() {
 	var fileIndex uint64
-	RegisterCLI("extract", 1, func(args []string) (uefi.Visitor, error) {
+	RegisterCLI("extract", "extract the files to a directory", 1, func(args []string) (uefi.Visitor, error) {
```
GanShun: You seem to have inconsistent capitalization on the beginning of the help string. should we standardize on lower case for everything?

GanShun:
Also, this is going to conflict with #155, so you'll need to fix it up when ryan merges 155

rminnich:
I'll wait but sure I can add it to usage

rjoleary:
#155 is now merged, you'll need to add help string to `count`.

pkg/visitors/extract.go
rminnich: yes

PR #162, opened 2018-09-14
pkg/uefi: fix to be more paranoid if a buffer is too short or a header malformed.

rminnich:
I had a malformed image created by utk which was causing this:
rminnich@uroot:~/projects/NERF/equus$ utk new.rom table
panic: runtime error: slice bounds out of range

goroutine 1 [running]:
github.com/linuxboot/fiano/pkg/uefi.NewFlashImage(0xc00009e000, 0x1a26000, 0x1a26200, 0x14, 0x0, 0x0)
	/home/rminnich/go/src/github.com/linuxboot/fiano/pkg/uefi/flash.go:249 +0xf99
github.com/linuxboot/fiano/pkg/uefi.Parse(0xc00009e000, 0x1a26000, 0x1a26200, 0x1a26000, 0x1a26200, 0x0, 0x0)
	/home/rminnich/go/src/github.com/linuxboot/fiano/pkg/uefi/uefi.go:133 +0xcc
main.main()
	/home/rminnich/go/src/github.com/linuxboot/fiano/cmds/utk/utk.go:109 +0x476

With this change we get this:

rminnich@uroot:~/projects/NERF/equus$ utk new.rom table
2018/09/13 19:52:49 27418624 byte buf is too short for end BIOS: need 33554432 bytes
rminnich@uroot:~/projects/NERF/equus$

The file new.rom was created by this:
utk test.rom remove '^Ip.*' save new.rom

Arguably this should also be caught by Validate, but if a programmer
makes a mistake and does not call Validate it's nice to have an error message.

I took the opportunity to change the code a bit so that there is a common
for loop over a [] of structs using a closure to isolate some of the operations.

Take a look and you'll see what I mean. I think overall it's probably better
as we can write the common checking code once, not 5 times.

### DIFF
--- a/pkg/uefi/flash.go
+++ b/pkg/uefi/flash.go
@@ -243,48 +243,70 @@ func NewFlashImage(buf []byte) (*FlashImage, error) {
 	f.regions = append(f.regions, &f.IFD)
 
 	// BIOS region
+	// We test this one case because it is not optional
 	if !f.IFD.Region.BIOS.Valid() {
 		return nil, fmt.Errorf("no BIOS region: invalid region parameters %v", f.IFD.Region.BIOS)
 	}
-	br, err := NewBIOSRegion(buf[f.IFD.Region.BIOS.BaseOffset():f.IFD.Region.BIOS.EndOffset()], &f.IFD.Region.BIOS)
-	if err != nil {
-		return nil, err
-	}
-	f.BIOS = br
-	// Add to extractable regions
-	f.regions = append(f.regions, f.BIOS)
 
-	// ME region
-	if f.IFD.Region.ME.Valid() {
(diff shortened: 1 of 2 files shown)


rminnich:
Note that I'm not going to fool with the test until it looks like the change is acceptable.

GanShun:
This change looks acceptable, but I want to know why the regions were wrong in the first place. 

Also I prefer the way you've done it here, and this should not be in Validate. As a rule of thumb, Validate contains errors that don't affect the parsing of the image. Anything that causes us to fail to parse the image further should fail in the constructor directly.

GanShun:
I looked more into this, and I'm ok with merging it, but what's actually happening is that there are now more regions beyond the 4 types and we're not handling it correctly, so upon reassembly we lose it.

rjoleary:
Looks acceptable to me! It will make it easier to add more regions in the future as well.

Now, just finish the test.

rjoleary:
@GanShun subsumed all these changes into #184

PR #165, opened 2018-09-14
Convert validate to a visitor

rjoleary:


pkg/uefi/file.go
```diff
@@ -178,23 +178,24 @@ func (a *fileAttr) setLarge(large bool) {
 }
 
 // Checks if we need to checksum the file body
```
rminnich: HasChecksum checks ...

pkg/uefi/file.go
```diff
@@ -157,7 +157,7 @@ type FileHeader struct {
 }
 
 // Checks if the large file attribute is set
```
rminnich: IsLarge checks ...

pkg/uefi/file.go
```diff
@@ -178,23 +178,24 @@ func (a *fileAttr) setLarge(large bool) {
...
 
 // Checks if we need to checksum the file body
-func (a fileAttr) hasChecksum() bool {
+func (a fileAttr) HasChecksum() bool {
 	return a&0x40 != 0
 }
 
 // HeaderLen is a helper function to return the length of the file header
 // depending on the file size
 func (f *File) HeaderLen() uint64 {
```
rminnich: I think you can drop 'is a helper function'
HeaderLen returns the length of the file header.

pkg/visitors/validate.go
```diff
@@ -0,0 +1,270 @@
...
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/linuxboot/fiano/pkg/uefi"
+)
+
+// Validate performs extra checks on the firmware image.
+type Validate struct {
+	// In
```
rminnich: hmm. Not sure I like these In and Out comments ....

pkg/uefi/file.go
rjoleary: I'll fix it, but it's true the linter has no problem, even with:

golint -min_confidence=0.0 file.go

PR #174, opened 2018-09-17
Use hexadecimal sizes in table visitor

rjoleary:
Additionally, only print the FV free space when the free space is not
empty.

### DIFF
--- a/pkg/visitors/table.go
+++ b/pkg/visitors/table.go
@@ -28,28 +28,28 @@ func (v *Table) Run(f uefi.Firmware) error {
 func (v *Table) Visit(f uefi.Firmware) error {
 	switch f := f.(type) {
 	case *uefi.FlashImage:
-		return v.printRow(f, "Image", "", "", "")
+		return v.printRow(f, "Image", "", "")
 	case *uefi.FirmwareVolume:
-		return v.printRow(f, "FV", f.FileSystemGUID.String(), "", f.Length)
+		return v.printRow(f, "FV", f.FileSystemGUID.String(), "")
 	case *uefi.File:
 		// TODO: make name part of the file node
-		return v.printRow(f, "File", f.Header.UUID.String(), f.Header.Type, f.Header.ExtendedSize)
+		return v.printRow(f, "File", f.Header.UUID.String(), f.Header.Type)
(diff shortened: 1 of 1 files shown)


rjoleary:
Fixes #168

pkg/visitors/table.go
```diff
@@ -28,50 +28,50 @@ func (v *Table) Run(f uefi.Firmware) error {
...
 
-func (v *Table) printRow(f uefi.Firmware, node, name, typez, size interface{}) error {
+func (v *Table) printRow(f uefi.Firmware, node, name, typez interface{}) error {
 	if v.W == nil {
 		v.W = tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
 		defer func() { v.W.Flush() }()
 		fmt.Fprintf(v.W, "%sNode\tGUID/Name\tType\tSize\n", indent(v.indent))
 	}
-	fmt.Fprintf(v.W, "%s%v\t%v\t%v\t%v\n", indent(v.indent), node, name, typez, size)
+	fmt.Fprintf(v.W, "%s%v\t%v\t%v\t%#8x\n", indent(v.indent), node, name, typez, len(f.Buf()))
```
GanShun: At some point we should make Size() be a function that checks the length of the buffer, and SetBuf reset the size.

pkg/visitors/table.go
```diff
@@ -28,50 +28,50 @@ func (v *Table) Run(f uefi.Firmware) error {
...
 	v2 := *v
 	v2.indent++
 	if err := f.ApplyChildren(&v2); err != nil {
 		return err
 	}
-	if fv, ok := f.(*uefi.FirmwareVolume); ok {
+	if fv, ok := f.(*uefi.FirmwareVolume); ok && fv.FreeSpace != 0 {
```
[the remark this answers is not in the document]
rjoleary: Done, I reverted it

rjoleary:
PTAL

pkg/visitors/table.go
GanShun: sort of. We need to also figure out how to always get the correct size from Length(), as what's underlying may have changed.

Issue #170, opened 2018-09-16
Rename UUID to GUID


rjoleary:
Found in various places of the code

rjoleary:
Let me know if you'd prefer the opposite: GUID => UUID. I'm happy as long as there is not a mix.

PR #175, opened 2018-09-17
Rename all instances of uuid to guid

rjoleary:
The UEFI spec refers to them as GUIDs.

### DIFF
--- a/pkg/guid/guid.go
+++ b/pkg/guid/guid.go
@@ -2,8 +2,8 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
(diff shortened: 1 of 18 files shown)


rjoleary:
Fixes #170

rjoleary:
Alternatively, let me know if they should all be GUID. Either way, let's not use a mixture.

rjoleary:
A funny quote from stackoverflow:

> The term GUID usually refers to Microsoft's implementation of the Universally Unique Identifier (UUID) standard.
> 
> Just like microsoft to take something that works everywhere (the universe) and make it work in a very small subset (the globe)
>
> https://stackoverflow.com/questions/246930/is-there-any-difference-between-a-guid-and-a-uuid

PR #183, opened 2018-09-17
Change the FV guid to be FFSV3 if there are large files

GanShun:

### DIFF
--- a/pkg/visitors/assemble.go
+++ b/pkg/visitors/assemble.go
@@ -17,6 +17,13 @@ import (
 
 // Assemble reconstitutes the firmware tree assuming that the leaf node buffers are accurate
 type Assemble struct {
+	// This is set when a file or section >=16MiB is encountered during assembly.
+	// This tells the enclosing FV to use the FFSV3 GUID instead of the FFSV2 GUID,
+	// and the enclosing FV resets it.
+	// TODO: figure out if, in the case where the FVs are triply nested, must the FVs further up
+	// also use the FFSV3 GUID? In that case we should fix this since only the innermost
+	// enclosing FV changes to FFSV3
+	useFFS3 bool
 }
 
 // Run just applies the visitor.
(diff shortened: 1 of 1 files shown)


GanShun:
Right now this just sets the enclosing FV to FFSV3, it is unclear if the FVs are even more multiply nested, does it apply to all the FVs higher up. I don't have an actual image of this form to test on, so I'm leaving it like this for now.

pkg/visitors/assemble.go
```diff
@@ -133,6 +134,15 @@ func (v *Assemble) Visit(f uefi.Firmware) error {
 		// TODO: handle the whole header instead of doing this
 		binary.LittleEndian.PutUint64(fBuf[32:], f.Length)
 
+		// Write the correct GUID to the correct spot
```
rjoleary: Can you add a note here saying to lookup the meaning of EFI_FIRMWARE_FILE_SYSTEM3_GUID in section 3.2.2, volume X in the PI spec version Y?

pkg/visitors/assemble.go
```diff
@@ -253,6 +267,9 @@ func (v *Assemble) Visit(f uefi.Firmware) error {
 
 		// Fix up the header
 		err = f.GenSecHeader()
+		if f.Header.ExtendedSize > 0xFFFFFF {
```
rjoleary: Is this where the spec is not clear? It doesn't say anything about sections >=16MiB, which is the case where a section can contain an encapsulated section which is compressed.

pkg/visitors/assemble.go
```diff
@@ -133,6 +134,15 @@ func (v *Assemble) Visit(f uefi.Firmware) error {
 		// TODO: handle the whole header instead of doing this
 		binary.LittleEndian.PutUint64(fBuf[32:], f.Length)
 
+		// Write the correct GUID to the correct spot
+		if v.UseFFS3 && f.FileSystemGUID == *uefi.FFS2 {
```
rjoleary: Is the opposite also true? `if !v.UseFFS3 && r.FileSystemGUID == *uefi.FFS3 { f.FilesystemGUID = *uefi.FFS2; ... }`

pkg/visitors/assemble.go
```diff
@@ -17,6 +17,7 @@ import (
 
 // Assemble reconstitutes the firmware tree assuming that the leaf node buffers are accurate
 type Assemble struct {
+	UseFFS3 bool
```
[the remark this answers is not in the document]
GanShun: Done

pkg/visitors/assemble.go
GanShun: The spec is wrong because it says this:
The EFI_COMMON_SECTION_HEADER2 structure is only used if the section is too large to
be described using EFI_COMMON_SECTION_HEADER . Large sections using
EFI_COMMON_SECTION_HEADER2 can only exist in a file using
EFI_FFS_FILE_HEADER2 , the FFS_ATTRIB_LARGE_FILE attribute in the file header is
set.

pkg/visitors/assemble.go
GanShun: I think it can be fine even if we don't downgrade. I would leave this until we see a case where we actually need to downgrade.

pkg/visitors/assemble.go
GanShun: Done

PR #195, opened 2018-09-26
Dump

GanShun:


pkg/visitors/dump.go
```diff
@@ -0,0 +1,72 @@
...
+	// The file is written to this writer.
+	W io.Writer
+}
+
+// Run just calls the visitor
+func (v *Dump) Run(f uefi.Firmware) error {
+	return f.Apply(v)
+}
+
+// Visit Uses find to dump a file to W
```
rjoleary: Need periods at the end of these sentences. Also "U" shouldn't be capitalized.

pkg/visitors/dump_test.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	// Read in expected file.
+	file, err := ioutil.ReadFile("../../integration/roms/testfile.ffs")
+	if err != nil {
+		t.Fatal(err)
+	}
+	// W should now contain the file.
+	if !bytes.Equal(b.Bytes(), file) {
+		// Should dump the file somewhere for comparison.
```
rjoleary: Prefix comment with "TODO".

PR #196, opened 2018-09-27
Insert Visitor

GanShun:
This adds 5 command line operations: insert_front, insert_end,
insert_in, insert_after, and insert_before

insert_front <FV GUID, File GUID, File Name> file.ffs:
Inserts at the front of the specified FV

insert_end <FV GUID, File GUID, File Name> file.ffs:
Inserts at the end of the specified FV

insert_in <FV GUID, File GUID, File Name> file.ffs:
Inserts somewhere in the specified FV

insert_after <File GUID, File Name> file.ffs:
Inserts after the specified file

insert_before <File GUID, File Name> file.ffs:
Inserts before the specified file

### DIFF
--- a/pkg/visitors/insert.go
+++ b/pkg/visitors/insert.go
@@ -1,0 +1,162 @@
+// Copyright 2018 the LinuxBoot Authors. All rights reserved
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package visitors
+
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+
+	"github.com/linuxboot/fiano/pkg/uefi"
+)
+
+// InsertType defines the insert type operation that is requested
+type InsertType int
+
+// Insert Types
+const (
+	// These first two specify a firmware volume.
+
+	// InsertFront inserts a file at the beginning of the firmware volume,
+	// which is specified by 1) FVname GUID, or (File GUID/File name) of a file
+	// inside that FV.
+	InsertFront InsertType = iota
+	// InsertEnd inserts a file at the end of the specified firmware volume.
+	InsertEnd
+
+	// These two specify a File to insert before or after
+	// InsertAfter inserts after the specified file,
+	// which is specified by a File GUID or File name.
+	InsertAfter
+	// InsertBefore inserts before the specified file.
+	InsertBefore
+	// TODO: Add InsertIn
+)
+
+var insertTypeNames = map[InsertType]string{
+	InsertFront:  "insert_front",
+	InsertEnd:    "insert_end",
+	InsertAfter:  "insert_after",
+	InsertBefore: "insert_before",
+}
+
+// String creates a string representation for the insert type.
+func (i InsertType) String() string {
+	if t, ok := insertTypeNames[i]; ok {
+		return t
+	}
+	return "UNKNOWN"
+}
+
+// Insert inserts a firmware file into an FV
+type Insert struct {
+	// Input
+	Predicate func(f uefi.Firmware) bool
+	NewFile   *uefi.File
+	InsertType
+
+	// Matched File
+	FileMatch uefi.Firmware
+}
+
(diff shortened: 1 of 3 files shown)


pkg/visitors/insert.go
```diff
@@ -0,0 +1,171 @@
...
+
+	// InsertFront inserts a file at the beginning of the firmware volume,
+	// which is specified by 1) FVname GUID, or  (File GUID/File name) of a file
+	// inside that FV.
+	InsertFront InsertType = iota
+	// InsertEnd inserts a file at the end of the specified firmware volume.
+	InsertEnd
+	// InsertIn inserts a file anywhere in the specified firmware volume.
+	// It works exactly like InsertEnd for now.
+	InsertIn
```
rjoleary: Should we still have `InsertIn` if it's not really implemented yet? People might depend on this behaviour.

pkg/visitors/insert_test.go
```diff
@@ -0,0 +1,146 @@
...
+		{InsertEnd.String(), InsertEnd},
+		{InsertIn.String(), InsertIn},
+		{InsertAfter.String(), InsertAfter},
+		{InsertBefore.String(), InsertBefore},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			f := parseImage(t)
+
+			file, err := ioutil.ReadFile("../../integration/roms/testfile.ffs")
```
rjoleary: Would it be a good idea to dedup the repetition in the tests?

pkg/visitors/insert.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	if numMatch := len(find.Matches); numMatch > 1 {
+		return fmt.Errorf("more than one match, only one match allowed! got %v", find.Matches)
+	} else if numMatch == 0 {
+		return errors.New("no matches found")
+	}
+
+	// Find should only match a file or a firmware volume. If it's an FV, we can
+	// edit the FV directly
```
rjoleary: Period at end of sentence.

rjoleary:
Looks very good!

pkg/visitors/insert.go
```diff
@@ -0,0 +1,171 @@
...
+
+// InsertType defines the insert type operation that is requested
+type InsertType int
+
+// Insert Types
+const (
+	// These first three specify a firmware volume.
+
+	// InsertFront inserts a file at the beginning of the firmware volume,
+	// which is specified by 1) FVname GUID, or  (File GUID/File name) of a file
```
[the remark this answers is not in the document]
GanShun: Done

pkg/visitors/insert.go
GanShun: What exactly would they depend on? that it inserts at the end? I could remove the comment and say that it's currently implemented like insertEnd but no one should depend on it

pkg/visitors/insert_test.go
GanShun: maybe, but I did this so that I would always get a new clean copy of the file. I could instead read once and do 5 copies. What do you think?

pkg/visitors/insert.go
GanShun: Done

pkg/visitors/insert_test.go
rjoleary: Sorry, no what I mean is the there's ~15 lines in each of these three functions which are the same:

```
			f := parseImage(t)
 			file, err := ioutil.ReadFile("../../integration/roms/testfile.ffs")
			if err != nil {
				t.Fatal(err)
			}
 			ffs, err := uefi.NewFile(file)
			if err != nil {
				t.Fatal(err)
			}
			// Apply the visitor.
			insert := &Insert{
				Predicate:  FindFileGUIDPredicate(*testGUID),
				NewFile:    ffs,
				InsertType: test.InsertType,
			}
```

It can be refactored into a function.

PR #198, opened 2018-10-01
Have ReadArea return a SectionReader

rminnich:
The SectionReader is a bit more flexible and makes
cbfs simpler.


rminnich:
never mind, I'll do this some other way, man I am so out of it today

GanShun:
I saw that you broke the interface yeah.. we really should eventually merge fmap functionality into utk

PR #211, opened 2018-10-30
Anchor regex for find and other commands

### LINKED ISSUE #202: Anchor Find Regex
The find regex should automatically be anchored with ^ and $ to avoid accidental matches.

rjoleary:
This follows the principle of least astonishment.

Fixes #202


pkg/visitors/find.go
```diff
@@ -96,7 +96,7 @@ func FindFileGUIDPredicate(r guid.GUID) func(f uefi.Firmware) bool {
 
 // FindFilePredicate is a generic predicate for searching files and UI sections only.
 func FindFilePredicate(r string) (func(f uefi.Firmware) bool, error) {
-	searchRE, err := regexp.Compile(r)
+	searchRE, err := regexp.Compile("^" + r + "$")
```
GanShun: what happens if someone manually anchors the regex?

GanShun: Ok that's fine.

Issue #217, opened 2018-11-02
Option to use system xz command


rjoleary:

rjoleary:
This is required to use utk in linuxboot/linuxboot.

PR #221, opened 2018-11-05
compression: Optionally use the system's xz cmd

rjoleary:
Using your system's xz command (not the default) will give better
compression when compared to the Go-based compressor. My xz
implementation also appears to be slightly faster.


rjoleary:
Fixes #217

pkg/compression/systemlzma.go
```diff
@@ -0,0 +1,66 @@
...
+	cmd.Stdin = bytes.NewBuffer(decodedData)
+	encodedData, err := cmd.Output()
+	if err != nil {
+		return nil, err
+	}
+
+	// Quoting the XZ Utils manpage:
+	//
+	//     xz supports decompressing .lzma files with or without
+	//     end-of-payload marker, but all.lzma files created by xz will use
```
GanShun: all .lzma

pkg/compression/x86.go
```diff
@@ -2,11 +2,23 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-package lzma
+package compression
 
-// DecodeX86 decodes LZMA data with the x86 extension.
-func DecodeX86(encodedData []byte) ([]byte, error) {
-	decodedData, err := Decode(encodedData)
+// LZMAX86 implements Compressor and includes X86 filter which is popular in
```
GanShun: includes *the* X86 filter

pkg/compression/compression.go
```diff
@@ -0,0 +1,53 @@
...
+func CompressorFromGUID(guid *guid.GUID) Compressor {
+	switch *guid {
+	case LZMAGUID:
+		if *xzPath != "" {
+			return &SystemLZMA{*xzPath}
+		}
+		return &LZMA{}
+	case LZMAX86GUID:
+		if *xzPath != "" {
+			// Alternatively, the -f86 argument could be passed
```
GanShun: The current method is probably a good way to make sure we can still decompress regular xz compressed stuff

Issue #181, opened 2018-09-17
Run visitor tests on ROMs other than OVMF


rjoleary:

rjoleary:
At the present the integration test performs the following:

1. Extract OVMF image to folder A
2. Assembly A to image X
3. Extract X to folder B
4. Assert `diff --recursive A B` == 0

The OVMF image and image X cannot be compared directly because they likely use different compression schemes. The integration test solves this problem by comparing the comparing the extracted (decompressed) files and their metadata.

This works fine for the OVMF image checked into the UTK source tree, but we found for some other images (such as the Tyan image [publicly available on their downloads page](https://www.tyan.com/Motherboards=S7106=S7106GM2NR=downloads=EN)) have a Firmware Volume which expands in size when the image is re-compressed in step 2. This causes the FV metadata and headers to differ between A and B. The difference causes the test to fail.

GanShun:
I like the hash idea! That's a good idea, we can hash only extracted file contents, skipping the headers.

PR #222, opened 2018-11-06
integration: Test JSON changes against golden

rjoleary:
We were unable to devise a satisfactory way to automatically test
extraction on entire images. This solution requires approximately a
minute of work to compare the new JSON output against the previous
JSON for each commit.

### DIFF
--- a/integration/utk_test.go
+++ b/integration/utk_test.go
@@ -5,13 +5,17 @@
 package utk_test
 
 import (
+	"bytes"
 	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 	"testing"
 
+	"github.com/linuxboot/fiano/pkg/uefi"
 	"github.com/linuxboot/fiano/pkg/utk"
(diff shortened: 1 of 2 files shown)


rjoleary:
For #181

integration/utk_test.go
```diff
@@ -101,3 +105,53 @@ func TestExtractAssembleExtract(t *testing.T) {
...
+//     utk integration/roms/OVMF.rom json > integration/roms/OVMF.json
+//
+// Otherwise, this test will fail. This gives you a chance to review how your
+// code affects the tree and identify any mistakes.
+func TestRegressionJson(t *testing.T) {
+	// Create a temporary directory.
+	tmpDir := createTempDir(t)
+	defer os.RemoveAll(tmpDir)
+
+	for _, tt := range romList(t) {
```
GanShun: how about you iterate over the jsonlist for this test? so we can have roms without the golden json, and this test only runs if you have those

rjoleary: Ok, sounds good!

PR #223, opened 2018-11-08
Add DXECleaner to automate removal of DXEs

rjoleary:
A unit test is included. I'll add example usage to a future commit or to
the LinuxBoot Book.

### DIFF
--- a/pkg/visitors/dxecleaner.go
+++ b/pkg/visitors/dxecleaner.go
@@ -1,0 +1,162 @@
+// Copyright 2018 the LinuxBoot Authors. All rights reserved
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package visitors
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"os/signal"
+	"path/filepath"
+	"syscall"
+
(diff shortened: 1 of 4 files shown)


pkg/visitors/find.go
```diff
@@ -94,6 +94,16 @@ func FindFileGUIDPredicate(r guid.GUID) func(f uefi.Firmware) bool {
 	}
 }
 
+// FindFileTypePredicate is a generic predicate for searching file types only.
+func FindFileTypePredicate(t uefi.FVFileType) func(f uefi.Firmware) bool {
```
GanShun: We want to remove more than one type at a time. It would be nice to instead specify entire firmware volumes to run the cleaner on. You don't have to change this for now. I'll file an issue

pkg/visitors/dxecleaner.go
```diff
@@ -0,0 +1,147 @@
...
+				if err != nil {
+					return true, err
+				}
+				defer os.RemoveAll(tmpDir)
+				tmpFile := filepath.Join(tmpDir, "bios.bin")
+
+				if err := (&Save{tmpFile}).Run(f); err != nil {
+					return true, err
+				}
+				if err := exec.CommandContext(ctx, args[0], tmpFile).Run(); err != nil {
```
GanShun: handle different kinds of failure
0: success
1: hard failure/tool failure
2: test failure

PR #231, opened 2018-11-21
Forth utk

rminnich:
This is a forth utk.

I think doing it as a new program has some benefits.
I notice we're wiring some assumptions into utk package for the
utk program, and it really becomes apparent when you try to
write a second program. This way, we can make sure the utk package
will work for more than just utk.

I no longer have to annoy Gan about the forth interface, and
I no longer have to annoy Ryan with commands I want :-)

This is a trivial interactive forth program, not really
done maybe, that lets me do a bit more messing about
a bit more easily than utk.

The model is that one can read in an fv, see things, mess
with things and, when done, push things out. You can make
decisions about what to remove, and then undo those
decisions, and only commmit when you type 'go'.

You can tag things with arbitrary tags and arbitrary text,
said text can even be output of other commands.

Two tags, r and k, have meaning to futk; r means 'removing when
writing firmware volume', and k means 'add to blacklist for
dxecleaner'.

The go command creates a new image. You can think of it as a commit.
Again, if you have tagged any DXEs as k, it writes out a blacklist suitable
(we hope) for a dxeclean pass; things tagged r are removed.

Here's an example session:

rminnich@uroot:~/go/src/github.com/linuxboot/fiano/cmds/futk$ ./futk 7106V100.ROM fv
2018/11/21 09:14:29 Found 413 things
413
OK Tcp|Pxe|Udp ix r r tag
[6D6963AB-906D-4A65-A7CA-BD40E5D6AF2B:EFI_FV_FILETYPE_DRIVER:Udp4Dxe D912C7BC-F098-4367-92BA-E911083C7B0E:EFI_FV_FILETYPE_DRIVER:Udp6Dxe B95E9FDA-26DE-48D2-8807-1F9107AC5E3A:EFI_FV_FILETYPE_DRIVER:UefiPxeBcDxe 1A7E4468-2F55-4A56-903C-01265EB7622B:EFI_FV_FILETYPE_DRIVER:TcpDxe]
OK Tcp|Pxe|Udp ix status
1A7E4468-2F55-4A56-903C-01265EB7622B:EFI_FV_FILETYPE_DRIVER:TcpDxe: res map[r:r]
6D6963AB-906D-4A65-A7CA-BD40E5D6AF2B:EFI_FV_FILETYPE_DRIVER:Udp4Dxe: res map[r:r]
D912C7BC-F098-4367-92BA-E911083C7B0E:EFI_FV_FILETYPE_DRIVER:Udp6Dxe: res map[r:r]
B95E9FDA-26DE-48D2-8807-1F9107AC5E3A:EFI_FV_FILETYPE_DRIVER:UefiPxeBcDxe: res map[r:r]

OK bzImage.uroot splat
[7C04A583-9E3E-4F1C-AD65-E05268D0B4D1:EFI_FV_FILETYPE_APPLICATION:FullShell]
OK go
[/tmp/futkblacklist753009682 /tmp/futkrom522877513]
OK

OK is the prompt.
In Forth fashion, OK is printed even if things are not, strictly speaking, OK.
On an old Sun, flames could be shooting out the top, and the firmware prompt
would still be
OK

Breaking this down:
Tcp|Pxe|Udp ix
will push an array on TOS of the things it finds. The
r r tag
uses that array and tags it with an r tag, set in a map; you can use any value
for the r tag you want. tag value is first, name second and,
note, the value can be a forth expression:
OK APPLI ix 2 2 + r tag
[7C04A583-9E3E-4F1C-AD65-E05268D0B4D1:EFI_FV_FILETYPE_APPLICATION:FullShell]
OK APPLI ix status
7C04A583-9E3E-4F1C-AD65-E05268D0B4D1:EFI_FV_FILETYPE_APPLICATION:FullShell: res map[r:4]

When the go command is run, things tagged with r are removed. The value of the r tag
is of no consequence.

Some DXEs you want to keep; you can do a similar thing with ix and tag
them as k.

GUIDs of DXEs tagged with k are written to the blacklist file.

The splat command does a transplat on the selected SET OF files that match .*Shell.*
So, in this case, we transplat these with bzImage.uroot. It shows us
the identities of the file(s) it replaced. It also adds a tag to the file, which is
ignored by the go command, but useful to humans:
OK Shell ix status
7C04A583-9E3E-4F1C-AD65-E05268D0B4D1:EFI_FV_FILETYPE_APPLICATION:FullShell: res map[replaced:shell]

Note that the name includes GUID, type, and name, so you can do stuff like this:
OK APPLICATION ix status
7C04A583-9E3E-4F1C-AD65-E05268D0B4D1:EFI_FV_FILETYPE_APPLICATION:FullShell: res map[]

i.e. show all applications.

Obviously there's lots that can be done but I'm already finding this easier to use than utk.

I think the combination of simple interactivity and the ability to backout and or finely add changes helps.

The utk visitor pattern made this all so incredibly easy, FYI.

If you keep misreading futk for Something Else, well, I think
it's more a comment on you than the name of the program :-)
So There! :-)

### DIFF
--- a/cmds/futk/futk.go
+++ b/cmds/futk/futk.go
@@ -1,0 +1,564 @@
+// Copyright 2018 the LinuxBoot Authors. All rights reserved
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// futk is a forth-inspired infterface to utk.
+// It is in a state of flux just now.
+// Typical usage, and note that our first command is in argv, and they could all (or none) be in argv:
+// rminnich@uroot:~/go/src/github.com/linuxboot/fiano/cmds/futk$ ./futk 7106V100.ROM fv # fv reads in a firmware volume
+//    2018/11/21 09:14:29 Found 413 things
+//    413
+//    OK Tcp|Pxe|Udp ix r r tag
+//    [6D6963AB-906D-4A65-A7CA-BD40E5D6AF2B:EFI_FV_FILETYPE_DRIVER:Udp4Dxe D912C7BC-F098-4367-92BA-E911083C7B0E:EFI_FV_FILETYPE_DRIVER:Udp6Dxe B95E9FDA-26DE-48D2-8807-1F9107AC5E3A:EFI_FV_FILETYPE_DRIVER:UefiPxeBcDxe 1A7E4468-2F55-4A56-903C-01265EB7622B:EFI_FV_FILETYPE_DRIVER:TcpDxe]
+//    OK Tcp|Pxe|Udp ix status
+//    1A7E4468-2F55-4A56-903C-01265EB7622B:EFI_FV_FILETYPE_DRIVER:TcpDxe: tags map[r:r]
+//    6D6963AB-906D-4A65-A7CA-BD40E5D6AF2B:EFI_FV_FILETYPE_DRIVER:Udp4Dxe: tags map[r:r]
+//    D912C7BC-F098-4367-92BA-E911083C7B0E:EFI_FV_FILETYPE_DRIVER:Udp6Dxe: tags map[r:r]
+//    B95E9FDA-26DE-48D2-8807-1F9107AC5E3A:EFI_FV_FILETYPE_DRIVER:UefiPxeBcDxe: tags map[r:r]
+//
+//    OK bzImage.uroot splat
+//    [7C04A583-9E3E-4F1C-AD65-E05268D0B4D1:EFI_FV_FILETYPE_APPLICATION:FullShell]
+//    OK go
+//    [/tmp/futkblacklist753009682 /tmp/futkrom522877513]
+//    OK
+// The r tags means those DXEs are removed before writing.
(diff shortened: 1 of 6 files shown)


cmds/futk/futk.go
```diff
@@ -0,0 +1,554 @@
...
+		uefi.FVFileTypeCombinedSMMDXE:     true,
+		uefi.FVFileTypeSMMCore:            true,
+		uefi.FVFileTypeSMMStandalone:      true,
+		uefi.FVFileTypeSMMCoreStandalone:  true,
+		uefi.FVFileTypeOEMMin:             true,
+		uefi.FVFileTypeOEMMax:             true,
+		uefi.FVFileTypeDebugMin:           true,
+		uefi.FVFileTypeDebugMax:           true,
+		uefi.FVFileTypePad:                true,
+		// uefi.FVFileTypeFFSMin:             true, same as Pad, bug?
```
GanShun: it's what the spec says

PR #236, opened 2019-01-03
Remove except

GanShun:
This is necessary for linuxboot/linuxboot#36. Tests are coming in a bit, but take a look and see what you think.

### DIFF
--- a/pkg/visitors/dxecleaner.go
+++ b/pkg/visitors/dxecleaner.go
@@ -121,16 +121,19 @@ func (v *DXECleaner) Visit(f uefi.Firmware) error {
 // Empty lines and lines beginning with '#' are ignored.
 func parseBlackList(fileName, fileContents string) (string, error) {
 	blackList := ""
-	for _, line := range strings.Split(fileContents, "\n") {
+	for i, line := range strings.Split(fileContents, "\n") {
 		line = strings.TrimSpace(line)
 		if line == "" || strings.HasPrefix(line, "#") {
 			continue
 		}
-		_, err := regexp.Compile(line)
+		// skip multiple words, We need this for linuxboot file parsing
+		fw := strings.Split(line, " ")[0]
+		_, err := regexp.Compile(fw)
 		if err != nil {
-			return "", fmt.Errorf("cannot compile regex %q from blacklist file %q: %v", line, fileName, err)
+			return "", fmt.Errorf("cannot compile regex %q from blacklist file %q on line %v: %v",
+				fw, fileName, i, err)
 		}
-		blackList += "|(" + line + ")"
+		blackList += "|(" + fw + ")"
 	}
 	if blackList != "" {
 		blackList = blackList[1:]
--- a/pkg/visitors/find.go
+++ b/pkg/visitors/find.go
@@ -113,10 +113,14 @@ func FindFilePredicate(r string) (func(f uefi.Firmware) bool, error) {
 	if err != nil {
 		return nil, err
 	}
+	ciRE, err := regexp.Compile("^(?i)" + r + "$")
+	if err != nil {
+		return nil, err
+	}
 	return func(f uefi.Firmware) bool {
 		switch f := f.(type) {
 		case *uefi.File:
-			return searchRE.MatchString(f.Header.GUID.String())
+			return ciRE.MatchString(f.Header.GUID.String())
 		case *uefi.Section:
 			return searchRE.MatchString(f.Name)
 		}
@@ -130,12 +134,16 @@ func FindFileFVPredicate(r string) (func(f uefi.Firmware) bool, error) {
 	if err != nil {
 		return nil, err
 	}
+	ciRE, err := regexp.Compile("^(?i)" + r + "$")
+	if err != nil {
+		return nil, err
+	}
 	return func(f uefi.Firmware) bool {
 		switch f := f.(type) {
 		case *uefi.FirmwareVolume:
 			return searchRE.MatchString(f.FVName.String())
 		case *uefi.File:
-			return searchRE.MatchString(f.Header.GUID.String())
+			return ciRE.MatchString(f.Header.GUID.String())
 		case *uefi.Section:
 			return searchRE.MatchString(f.Name)
 		}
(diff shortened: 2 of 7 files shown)


GanShun:
Very nice, ineffassign caught my error

pkg/visitors/dxecleaner.go
```diff
@@ -126,11 +126,13 @@ func parseBlackList(fileName, fileContents string) (string, error) {
 		if line == "" || strings.HasPrefix(line, "#") {
 			continue
 		}
-		_, err := regexp.Compile(line)
+		// skip multiple words, We need this for linuxboot file parsing
+		fw := strings.Split(line, " ")[0]
+		_, err := regexp.Compile(fw)
 		if err != nil {
-			return "", fmt.Errorf("cannot compile regex %q from blacklist file %q: %v", line, fileName, err)
+			return "", fmt.Errorf("cannot compile regex %q from blacklist file %q: %v", fw, fileName, err)
```
rminnich: consider printing out the line here. In the for, for i, line := range
then you can print out the line number, the line, and the first element. It's helpful.

pkg/visitors/find.go
```diff
@@ -113,10 +113,14 @@ func FindFilePredicate(r string) (func(f uefi.Firmware) bool, error) {
 	if err != nil {
 		return nil, err
 	}
+	insenRE, err := regexp.Compile("^(?i)" + r + "$")
```
rminnich: or ciRE for case insensitive regexp? not important.

pkg/visitors/find.go
```diff
@@ -157,6 +165,58 @@ func FindAndPredicate(predicate1 FindPredicate, predicate2 FindPredicate) FindPr
...
+		return nil, fmt.Errorf("expected exactly one match, got %v, matches were: %v", mlen, find.Matches)
+	}
+	return find.Matches[0], nil
+}
+
+// FindEnclosingFV finds the FV that contains a file.
+func FindEnclosingFV(f uefi.Firmware, file *uefi.File) (*uefi.FirmwareVolume, error) {
+	pred := func(f uefi.Firmware) bool {
+		switch f := f.(type) {
+		case *uefi.FirmwareVolume:
```
rminnich: you could even call this 
DotDot

ha ha

pkg/visitors/find.go
```diff
@@ -157,6 +165,58 @@ func FindAndPredicate(predicate1 FindPredicate, predicate2 FindPredicate) FindPr
...
+			}
+		}
+		return false
+	}
+	dxeFV, err := FindExactlyOne(f, pred)
+	if err != nil {
+		return nil, fmt.Errorf("unable to find DXE FV, got: %v", err)
+	}
+	// result must be a FV.
+	fv := dxeFV.(*uefi.FirmwareVolume)
```
rminnich: could you do this in a way that won't panic on a type assertion? You could do a type switch:
switch v := dxeFV.(type) {
case *uefi.FirmwareVolume:
default:
fmt.Errorf("%v: got %v, want *uefi.FirmwareVolume", dxeFV, v)
}

I suspect this will be better in the long run.

pkg/visitors/find.go
```diff
@@ -157,6 +165,58 @@ func FindAndPredicate(predicate1 FindPredicate, predicate2 FindPredicate) FindPr
...
+	// We identify the Dxe Firmware Volume via the presence of the DxeCore
+	// This will cause problems if there are multiple dxe volumes.
+	pred := FindFileTypePredicate(uefi.FVFileTypeDXECore)
+	dxeCore, err := FindExactlyOne(f, pred)
+	if err != nil {
+		return nil, fmt.Errorf("unable to find DXE Core, got: %v", err)
+	}
+
+	// result must be a File.
+	file := dxeCore.(*uefi.File)
```
rminnich: same here.

pkg/visitors/remove.go
```diff
@@ -5,14 +5,18 @@
...
 )
 
 // Remove all firmware files with the given GUID.
 type Remove struct {
 	// Input
-	Predicate func(f uefi.Firmware) bool
-	Pad       bool
+	Predicate  func(f uefi.Firmware) bool
+	Pad        bool
+	RemoveDxes bool // I hate this, but there's no good way to work around our current structure
```
rminnich: I still prefer how I did it in futk but you know that already.

The whole blacklist file is a poor fit with the rest of the nice package.

pkg/visitors/remove.go
```diff
@@ -26,6 +30,37 @@ func (v *Remove) Run(f uefi.Firmware) error {
 	find := Find{
 		Predicate: v.Predicate,
 	}
+	if v.RemoveDxes {
```
rminnich: If you had a slice of things to remove, and that slice could be empty, then this could just be
for _, v := range stuffToRemove {

this bool feels a bit unneeded.

rminnich: But I admit I don't really get this code.

pkg/visitors/find.go
GanShun: I actually had a check in there, but I felt it wasn't necessary, since we have a predicate that can only ever match FVs. I figured if it ever changed enough that we suddenly weren't getting that back a panic was warranted.

pkg/visitors/remove.go
GanShun: it's worse than that. Matches is already the slice of things to remove. RemoveDxes is required to have logic so that we operate on the Dxe FV instead of the whole flash image. It's one area where forth would be useful. I cant get the cli to pass in the FV instead of the flash image, so I have to do it in Run().

GanShun: This is one case where I would agree with you that utk's cli interface has negatively affected the way I wrote this visitor.

pkg/visitors/dxecleaner.go
GanShun: Good idea, Thanks!

GanShun: Done

pkg/visitors/find.go
GanShun: Done

pkg/visitors/find.go
GanShun: Done

pkg/visitors/find.go
GanShun: Done

GanShun:
Tests are in

PR #240, opened 2019-01-27
nvram NVAR store parsing

JulienVdG:
Implement nvram NVAR store parsing

highly inspired from https://github.com/LongSoft/UEFITool/blob/new_engine/common/nvramparser.cpp

### DIFF
--- a/pkg/uefi/file.go
+++ b/pkg/uefi/file.go
@@ -218,9 +218,12 @@ type FileHeaderExtended struct {
 
 // File represents an EFI File.
 type File struct {
-	Header   FileHeaderExtended
-	Type     string
-	Sections []*Section `json:",omitempty"`
+	Header FileHeaderExtended
+	Type   string
+
+	// a File can contain either Sections or an NVarStore but not both
+	Sections  []*Section `json:",omitempty"`
+	NVarStore *NVarStore `json:",omitempty"`
 
 	//Metadata for extraction and recovery
 	buf         []byte
@@ -410,6 +413,15 @@ func NewFile(buf []byte) (*File, error) {
 	newBuf := buf[:f.Header.ExtendedSize]
 	f.buf = make([]byte, f.Header.ExtendedSize)
 	copy(f.buf, newBuf)
+
+	// Special case for NVAR Store stored in raw file
+	if f.Header.Type == FVFileTypeRaw && f.Header.GUID == *NVAR {
+		ns, err := NewNVarStore(f.buf[f.DataOffset:])
+		if err != nil {
+			return nil, fmt.Errorf("error parsing NVAR store in file %v: %v", f.Header.GUID, err)
+		}
+		f.NVarStore = ns
+	}
 
 	// Parse sections
 	if _, ok := SupportedFiles[f.Header.Type]; !ok {
--- a/pkg/uefi/nvram.go
+++ b/pkg/uefi/nvram.go
@@ -11,6 +11,7 @@ package uefi
 
 import (
 	"bytes"
+	"crypto/sha256"
 	"encoding/binary"
 	"fmt"
 	"io"
@@ -77,6 +78,17 @@ func (t NVarEntryType) String() string {
 	return "UNKNOWN"
 }
 
+// NVarExtAttribute represent extended attributes
+type NVarExtAttribute uint8
+
+// Extended attributes values
+const (
+	NVarEntryExtChecksum    NVarExtAttribute = 0x01
+	NVarEntryExtAuthWrite   NVarExtAttribute = 0x10
+	NVarEntryExtTimeBased   NVarExtAttribute = 0x20
+	NVarEntryExtUnknownMask NVarExtAttribute = 0xCE
+)
+
 // NVar represent an NVAR entry
 type NVar struct {
 	Header    NVarHeader
@@ -91,9 +103,18 @@ type NVar struct {
 	Offset     uint64
 	NextOffset uint64
 
+	//Extended Header
+	ExtAttributes               *NVarExtAttribute `json:",omitempty"`
+	Checksum                    *uint8            `json:",omitempty"`
+	ExpectedChecksum            *uint8            `json:",omitempty"`
+	TimeStamp                   *uint64           `json:",omitempty"`
+	Hash                        []byte            `json:",omitempty"`
+	UnknownExtendedHeaderFormat bool              `json:",omitempty"`
+
 	//Metadata for extraction and recovery
 	buf        []byte
 	DataOffset int64
+	ExtOffset  int64 `json:",omitempty"`
 }
 
 // NVarStore represent an NVAR store
@@ -189,6 +210,87 @@ func (v *NVar) parseNext() error {
 		v.Type = LinkNVarEntry
 		v.NextOffset = v.Offset + next
 	}
+	return nil
+}
+
+func (v *NVar) parseExtendedHeader() error {
+	var knownExtDataFormat bool
+	if v.Header.Attributes&NVarEntryExtHeader == 0 {
+		return nil
+	}
+	var extendedHeaderSize uint16
+	r := bytes.NewReader(v.buf)
+	if _, err := r.Seek(-int64(binary.Size(extendedHeaderSize)), io.SeekEnd); err != nil {
+		return err
+	}
+	if err := binary.Read(r, binary.LittleEndian, &extendedHeaderSize); err != nil {
+		return err
+	}
+	// Sanity check extendedHeaderSize < body size
+	bodySize := int64(v.Header.Size) - v.DataOffset
+	if int64(extendedHeaderSize) > bodySize {
+		return fmt.Errorf("extended header size (%#x) is greater than body size (%#x)", extendedHeaderSize, bodySize)
+	}
+	v.ExtOffset = int64(v.Header.Size) - int64(extendedHeaderSize)
+	if _, err := r.Seek(v.ExtOffset, io.SeekStart); err != nil {
+		return err
+	}
+	var extAttributes NVarExtAttribute
+	if err := binary.Read(r, binary.LittleEndian, &extAttributes); err != nil {
+		return err
+	}
+	v.ExtAttributes = &extAttributes
+	// Variable with checksum
+	if extAttributes&NVarEntryExtChecksum != 0 {
+		// Get stored checksum
+		var storedChecksum uint8
+		storedChecksum = v.buf[int64(v.Header.Size)-int64(binary.Size(extendedHeaderSize)+binary.Size(storedChecksum))]
+		v.Checksum = &storedChecksum
+		// Recalculate checksum for the variable
+		calculatedChecksum := uint8(0)
+		// [0-3] _Skip_  entry 'NVAR' signature
+		// [4-5] Include entry size and flags
+		// [6-8] _Skip_  entry next (So linking will not invalidate the sum)
+		// [ 9 ] Include entry attributes
+		// [10-] Include entry data
+		for i := int64(4); i < int64(v.Header.Size); i++ {
+			calculatedChecksum += v.buf[i]
+			if i == 5 {
+				i += 3 // Skip Next
+			}
+		}
+		if calculatedChecksum != 0 {
+			calculatedChecksum = -calculatedChecksum
+			v.ExpectedChecksum = &calculatedChecksum
+		}
+		knownExtDataFormat = true
+	}
+
+	if v.Header.Attributes&NVarEntryAuthWrite == 0 {
+		var timestamp uint64
+		if err := binary.Read(r, binary.LittleEndian, &timestamp); err != nil {
+			switch err {
+			case io.EOF, io.ErrUnexpectedEOF:
+				return fmt.Errorf("extended header size (%#x) is too small for timestamp", extendedHeaderSize)
+			default:
+				return err
+			}
+		}
+		if v.Header.Attributes&NVarEntryDataOnly != 0 {
+			// Full or link variable have hash
+			hashstart, err := r.Seek(0, io.SeekCurrent)
+			if err != nil {
+				return err
+			}
+			if hashstart+sha256.Size > int64(v.Header.Size) {
+				return fmt.Errorf("extended header size (%#x) is too small for hash", extendedHeaderSize)
+			}
+			v.Hash = make([]byte, sha256.Size)
+			copy(v.Hash, v.buf[hashstart:hashstart+sha256.Size])
+		}
+		knownExtDataFormat = true
+	}
+	v.UnknownExtendedHeaderFormat = !knownExtDataFormat
 	return nil
 }
 
@@ -312,6 +414,13 @@ func newNVar(buf []byte, offset uint64, s *NVarStore) (*NVar, error) {
 	// Parse next node information
 	if err := v.parseNext(); err != nil {
 		return nil, err
+	}
+
+	// Entry with extended header
+	if err := v.parseExtendedHeader(); err != nil {
+		v.Name = fmt.Sprintf("Invalid ExtHeader, %v", err)
+		v.Type = InvalidNVarEntry
+		return &v, nil
 	}
 
 	// Entry is data-only (nameless and GUIDless entry or link)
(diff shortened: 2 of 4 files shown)


pkg/uefi/file.go
```diff
@@ -104,8 +104,9 @@ func (f FVFileType) String() string {
 
 // Stock GUIDS
 var (
-	ZeroGUID = guid.MustParse("00000000-0000-0000-0000-000000000000")
-	FFGUID   = guid.MustParse("FFFFFFFF-FFFF-FFFF-FFFF-FFFFFFFFFFFF")
+	ZeroGUID      = guid.MustParse("00000000-0000-0000-0000-000000000000")
+	FFGUID        = guid.MustParse("FFFFFFFF-FFFF-FFFF-FFFF-FFFFFFFFFFFF")
+	NVARStoreGUID = guid.MustParse("CEF5B9A3-476D-497F-9FDC-E98143E0422C")
```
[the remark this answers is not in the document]
JulienVdG: Reusing the already declared GUID "NVAR" instead of adding one.
Still the tested image has it in a raw file not in a fv.

pkg/uefi/file.go
```diff
@@ -218,9 +218,10 @@ type FileHeaderExtended struct {
 
 // File represents an EFI File.
 type File struct {
-	Header   FileHeaderExtended
-	Type     string
-	Sections []*Section `json:",omitempty"`
+	Header    FileHeaderExtended
+	Type      string
+	Sections  []*Section `json:",omitempty"`
```
rjoleary: Can you put a comment in front of this saying files either have `Sections` or `NVarStore`, but not both?

pkg/uefi/nvram.go
```diff
@@ -0,0 +1,475 @@
...
+
+	"github.com/linuxboot/fiano/pkg/guid"
+)
+
+// NVarAttribute represent Attributes
+type NVarAttribute uint8
+
+// Attributes
+const (
+	NVarEntryRuntime       NVarAttribute = 0x01
```
rjoleary: Is there a spec or reference implementation where this is defined, or has the whole thing been reverse engineered?

pkg/uefi/nvram.go
```diff
@@ -0,0 +1,325 @@
...
+		for {
+			if err := binary.Read(r, binary.LittleEndian, u); err != nil {
+				return err
+			}
+			if u[0] == 0 {
+				break
+			}
+			ucs2Name = append(ucs2Name, u[0])
+		}
+		v.Name = string(utf16.Decode(ucs2Name))
```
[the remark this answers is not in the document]
JulienVdG: Thanks, done

pkg/uefi/uefi.go
```diff
@@ -216,3 +216,13 @@ func Erase(buf []byte, polarity byte) {
 		buf[j] = Attributes.ErasePolarity
 	}
 }
+
+// IsErased check if the buffer is ErasePolarity
+func IsErased(buf []byte, polarity byte) bool {
+	for j, blen := 0, len(buf); j < blen; j++ {
```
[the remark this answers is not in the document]
JulienVdG: OK, I did just copy the for loop from the above function Erase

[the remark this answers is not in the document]
JulienVdG: Also both function have an unused 'polarity' parameter, shall I make another PR to get rid of it ?

pkg/uefi/file.go
JulienVdG: done

pkg/uefi/nvram.go
JulienVdG: Trammel Hudson have the same here: https://trmm.net/UEFI_notes
chipsec also: https://github.com/chipsec/chipsec/blob/master/chipsec/hal/uefi_platform.py#L143

JulienVdG: Probably found a reference to the source, it's definitely reverse engineered
https://firmwaresecurity.com/2016/04/16/nikolaj-on-nvram-formats-part-3/
Pointing to the blog of Nikolaj Schlej in russian:
https://habr.com/en/post/281901/

rjoleary: Thanks for digging up these links!

pkg/uefi/nvram.go
```diff
@@ -0,0 +1,458 @@
...
+	if _, err := r.Seek(-int64(binary.Size(extendedHeaderSize)), io.SeekEnd); err != nil {
+		return err
+	}
+	if err := binary.Read(r, binary.LittleEndian, &extendedHeaderSize); err != nil {
+		return err
+	}
+	// Sanity check extendedHeaderSize < body size
+	bodySize := int64(v.Header.Size) - v.DataOffset
+	if int64(extendedHeaderSize) > bodySize {
+		return fmt.Errorf("extended header size (0x%X) is greater than body size (0x%X)", extendedHeaderSize, bodySize)
```
rjoleary:
```suggestion
		return fmt.Errorf("extended header size (%#x) is greater than body size (%#x)", extendedHeaderSize, bodySize)
```

pkg/uefi/nvram.go
```diff
@@ -0,0 +1,458 @@
...
+		}
+		knownExtDataFormat = true
+	}
+
+	if v.Header.Attributes&NVarEntryAuthWrite == 0 {
+		var timestamp uint64
+		if err := binary.Read(r, binary.LittleEndian, &timestamp); err != nil {
+			switch err {
+			case io.EOF, io.ErrUnexpectedEOF:
+				return fmt.Errorf("extended header size (0x%X) is too small for timestamp", extendedHeaderSize)
```
rjoleary:
```suggestion
				return fmt.Errorf("extended header size (%#x) is too small for timestamp", extendedHeaderSize)
```

pkg/uefi/nvram.go
```diff
@@ -0,0 +1,458 @@
...
+	// Copy out the buffer.
+	s.buf = make([]byte, len(buf))
+	copy(s.buf, buf)
+
+	end := uint64(len(buf))
+
+	for offset := uint64(0); offset < end; {
+		v, err := newNVar(s.buf[offset:end], offset, &s)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing NVAR entry at offset 0x%x: %v", offset, err)
```
rjoleary:
```suggestion
			return nil, fmt.Errorf("error parsing NVAR entry at offset %#x: %v", offset, err)
```

pkg/uefi/nvram.go
```diff
@@ -0,0 +1,458 @@
...
+	}
+	ns, err := NewNVarStore(buf)
+	if err != nil {
+		return fmt.Errorf("error parsing NVAR store in var %v: %v", v.Name, err)
+	}
+	v.NVarStore = ns
+	return nil
+}
+
+// newNVar parses a sequence of bytes and returns a NVar
```
rjoleary:
```suggestion
// newNVar parses a sequence of bytes and returns an NVar
```

pkg/uefi/nvram.go
```diff
@@ -0,0 +1,458 @@
...
+// object, if a valid one is passed, returns nil if buf is clear, or an error.
+func newNVar(buf []byte, offset uint64, s *NVarStore) (*NVar, error) {
+	v := NVar{Type: FullNVarEntry, Offset: offset}
+	// read the header and check for existing NVAR
+	if err := v.parseHeader(buf); err != nil {
+		// No NVAR check if remaining space is erased
+		if IsErased(buf, Attributes.ErasePolarity) {
```
rjoleary: Can you move the `IsErased` check to the very beginning of this function?

pkg/uefi/nvram.go
```diff
@@ -0,0 +1,458 @@
...
+		if v.Header.Attributes&NVarEntryDataOnly != 0 {
+			// Full or link variable have hash
+			hashstart, err := r.Seek(0, io.SeekCurrent)
+			if err != nil {
+				return err
+			}
+			if hashstart+sha256.Size > int64(v.Header.Size) {
+				return fmt.Errorf("extended header size (0x%X) is too small for hash", extendedHeaderSize)
+			}
+			v.Hash = v.buf[hashstart : hashstart+sha256.Size]
```
GanShun: Suggest that you do a copy here in case the buffer is ever modified.

pkg/uefi/nvram.go
```diff
@@ -0,0 +1,458 @@
...
+	return nil
+}
+
+func (v *NVar) parseDataOnly(s *NVarStore) bool {
+	if v.Header.Attributes&NVarEntryDataOnly == 0 {
+		return false
+	}
+	// Search previously added entries for a link to this variable
+	var link *NVar
+	for _, l := range s.Entries {
```
GanShun: Are entries in order?  Is it possible to get this scenario

A links to B, which links to C, which has the data
but the entries are in the order
B, A, C?

pkg/uefi/nvram.go
```diff
@@ -0,0 +1,458 @@
...
+	return nil
+}
+
+func (s *NVarStore) getGUIDFromStore(i uint8) guid.GUID {
+	var GUID guid.GUID
+	if s.GUIDStoreLen < i+1 {
+		s.GUIDStoreLen = i + 1
+	}
+	r := bytes.NewReader(s.buf)
+	if _, err := r.Seek(-int64(binary.Size(GUID))*int64(i+1), io.SeekEnd); err != nil {
```
GanShun: There are some cases, such as extract where the standard behaviour is to not save as a binary what can be stored into the JSON. This means that the buffer won't exist when you first call ParseDir. I strongly suggest instead of relying on the presence of the buffer, directly parse all of the GUIDs at the point of parsing the NVAR store instead of relying on the existence of a buffer.

GanShun: In this case, this function would instead lookup the slice of GUIDs directly rather than seeking from the buffer.

pkg/uefi/nvram.go
JulienVdG: done

pkg/uefi/nvram.go
```diff
@@ -0,0 +1,458 @@
...
+	// read the header and check for existing NVAR
+	if err := v.parseHeader(buf); err != nil {
+		// No NVAR check if remaining space is erased
+		if IsErased(buf, Attributes.ErasePolarity) {
+			return nil, nil
+		}
+		return nil, err
+	}
+
+	//fmt.Printf("JVDG: len:%v h.s:%v s(h):%v\n", len(buf), v.Header.Size, binary.Size(v.Header))
```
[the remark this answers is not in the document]
JulienVdG: done

pkg/uefi/nvram.go
```diff
@@ -0,0 +1,458 @@
...
+			}
+		}
+		if v.Header.Attributes&NVarEntryDataOnly != 0 {
+			// Full or link variable have hash
+			hashstart, err := r.Seek(0, io.SeekCurrent)
+			if err != nil {
+				return err
+			}
+			if hashstart+sha256.Size > int64(v.Header.Size) {
+				return fmt.Errorf("extended header size (0x%X) is too small for hash", extendedHeaderSize)
```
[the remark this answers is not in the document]
JulienVdG: done

pkg/uefi/nvram.go
JulienVdG: done

pkg/uefi/nvram.go
JulienVdG: Right, in my mind it was more efficient to follow the list of headers instead of checking the full buffer for FFs, but you are right `IsErased` will stop really early if `buf` is not empty... so will do

pkg/uefi/nvram.go
```diff
@@ -0,0 +1,458 @@
...
+		// Get stored checksum
+		var storedChecksum uint8
+		storedChecksum = v.buf[int64(v.Header.Size)-int64(binary.Size(extendedHeaderSize)+binary.Size(storedChecksum))]
+		v.Checksum = &storedChecksum
+		// Recalculate checksum for the variable
+		calculatedChecksum := uint8(0)
+		// [4-5] Include entry size and flags
+		// [ 9 ] Include entry attributes
+		// [10-] Include entry data
+		for i := int64(4); i < int64(v.Header.Size); i++ {
```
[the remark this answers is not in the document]
JulienVdG: I don't have test data so this is Nikolaj's code ported to go...
Maybe I should leave the extended header parsing out of this PR until I can test it ?

pkg/uefi/nvram.go
```diff
@@ -0,0 +1,458 @@
...
+			calculatedChecksum += v.buf[i]
+			if i == 5 {
+				i += 3 // Skip Next
+			}
+		}
+		if calculatedChecksum != 0 {
+			calculatedChecksum = -calculatedChecksum
+			v.ExpectedChecksum = &calculatedChecksum
```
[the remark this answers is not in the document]
JulienVdG: Not really, the value will be present in JSON only if it's not valid... more an information.

pkg/uefi/nvram.go
JulienVdG: done

pkg/uefi/nvram.go
JulienVdG: It's reverse engineered, so who knows ! :man_shrugging:

But links are used to replace the value of a var without repeating the GUID and Name so I don't expect out of order entries.

pkg/uefi/nvram.go
JulienVdG: Yes seen that when implementing `ParseDir`.
The thing to take care is that GUID are stored in reverse order and we only know how much GUID are stored when the parsing is done.
I'm going rework that.

JulienVdG:
Thanks for the review !

Except for the technical part, what's needed ?
Do you want me to squash the commits (at least some of them ? the extended header is untested and should IMHO stay a separate commit)
To add signoff ?

Anything else ?

pkg/uefi/nvram.go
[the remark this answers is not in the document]
GanShun: This is interesting. Is there no extended header in any of winterfell's NVRAM variables? We can leave it, I would suggest you just leave a comment explicitly stating what's skipped

pkg/uefi/nvram.go
GanShun: Seems reasonable, just add a detailed comment here so we know to watch out for it if we ever start seeing weird NVRAM link orders.

pkg/uefi/nvram.go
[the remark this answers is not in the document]
GanShun: I'm ok with this for now.

If you look in the rest of the package, the checksum validation for the other fields are done in the validate visitor, and not done while parsing. This was done a long time ago because we had some rom which failed checksums while parsing. I'm now no longer sure that's the right decision, and maybe we should do aggressive checksum checking during parsing and emit error messages or warnings if we see them, because if the file header is corrupt, who's to know if the size we read off the header is actually correct?

Long term I would rather remove the checksum calculations from validate.go, and instead emit warnings or errors. Perhaps it should default to an error, and take a flag to turn those to warnings.

pkg/uefi/nvram.go
GanShun: Thank you!

GanShun:
oh please add a signoff, you can squash the commits and keep the extended header stuff separate.

pkg/uefi/uefi.go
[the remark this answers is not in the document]
JulienVdG: I'll remember this, resolving

pkg/uefi/nvram.go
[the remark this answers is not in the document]
JulienVdG: The validate visitor still currently fail on my winterfell images...
So while I agree with your long term goal, some more work will probably be needed before that... I guess some stuff are parsed while some other flags would have told upfront that the portion is invalid and should not be parsed ... (really just a guess)

pkg/uefi/nvram.go
JulienVdG: added

pkg/uefi/nvram.go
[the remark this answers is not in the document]
JulienVdG: updated the comment

pkg/uefi/nvram.go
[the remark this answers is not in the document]
JulienVdG: I put that on my TODO list

JulienVdG:
Sorry, I missed a few comments yesterday :(
Should be good now !

PR #246, opened 2019-01-30
Assemble depex

GanShun:


pkg/uefi/section.go
```diff
@@ -170,6 +170,16 @@ var DepExOpCodes = map[byte]DepExOpCode{
 	0x9: "SOR",
 }
 
+// DepExNamesToOpCodes maps the operation back to the code.
+var DepExNamesToOpCodes map[DepExOpCode]byte
+
+func init() {
+	DepExNamesToOpCodes = make(map[DepExOpCode]byte)
```
rjoleary: Delete this line

pkg/uefi/section.go
```diff
@@ -170,6 +170,16 @@ var DepExOpCodes = map[byte]DepExOpCode{
 	0x9: "SOR",
 }
 
+// DepExNamesToOpCodes maps the operation back to the code.
+var DepExNamesToOpCodes map[DepExOpCode]byte
```
rjoleary:
```suggestion
var DepExNamesToOpCodes = map[DepExOpCode]byte{}
```

pkg/visitors/assemble_test.go
```diff
@@ -0,0 +1,48 @@
+// Copyright 2018 the LinuxBoot Authors. All rights reserved
```
rjoleary: 2019

pkg/visitors/assemble.go
```diff
@@ -242,6 +242,35 @@ func (v *Assemble) Visit(f uefi.Firmware) error {
...
+					opcode, ok := uefi.DepExNamesToOpCodes[d.OpCode]
+					if !ok {
+						return fmt.Errorf("unable to map depex opcode string to opcode, string was: %v",
+							d.OpCode)
+					}
+					newBuf = append(newBuf, opcode)
+					// Sanity checks.
+					switch d.OpCode {
+					case "BEFORE", "AFTER", "PUSH":
+						// GUID should not be nil.
```
rjoleary:
```suggestion
						// GUID must not be nil.
```

pkg/visitors/assemble.go
```diff
@@ -242,6 +242,35 @@ func (v *Assemble) Visit(f uefi.Firmware) error {
...
+						if d.GUID == nil {
+							return fmt.Errorf("depex opcode %v should not have nil guid",
+								d.OpCode)
+						}
+						newBuf = append(newBuf, d.GUID[:]...)
+					default:
+						// GUID should be nil.
```
rjoleary:
```suggestion
						// GUID must be nil.
```

pkg/visitors/assemble.go
```diff
@@ -242,6 +242,35 @@ func (v *Assemble) Visit(f uefi.Firmware) error {
...
+						return fmt.Errorf("unable to map depex opcode string to opcode, string was: %v",
+							d.OpCode)
+					}
+					newBuf = append(newBuf, opcode)
+					// Sanity checks.
+					switch d.OpCode {
+					case "BEFORE", "AFTER", "PUSH":
+						// GUID should not be nil.
+						if d.GUID == nil {
+							return fmt.Errorf("depex opcode %v should not have nil guid",
```
rjoleary:
```suggestion
							return fmt.Errorf("depex opcode %v must not have nil guid",
```

pkg/visitors/assemble.go
```diff
@@ -242,6 +242,35 @@ func (v *Assemble) Visit(f uefi.Firmware) error {
...
+								d.OpCode)
+						}
+						newBuf = append(newBuf, d.GUID[:]...)
+					default:
+						// GUID should be nil.
+						if d.GUID != nil {
+							return fmt.Errorf("depex opcode %v should not have a guid! got %v",
```
rjoleary:
```suggestion
							return fmt.Errorf("depex opcode %v must not have a guid! got %v",
```

pkg/visitors/assemble.go
```diff
@@ -242,6 +242,35 @@ func (v *Assemble) Visit(f uefi.Firmware) error {
 				binary.LittleEndian.PutUint16(newBuf, f.BuildNumber)
 				newBuf = append(newBuf, unicode.UTF8ToUCS2(f.Version)...)
 				f.SetBuf(newBuf)
+			case uefi.SectionTypeDXEDepEx, uefi.SectionTypePEIDepEx,
+				uefi.SectionMMDepEx:
```
rjoleary: There are a few more sanity checks you could do, but I'm not sure if they are worth it. For example, PEIDepEx does not support all the opcodes. Also the PI spec says that "NoteThe PEI Foundation will only support the evaluation of dependency expressions that are less than or equal to 256 terms"

GanShun: I'll add a TODO

PR #247, opened 2019-01-31
Add SetErasePolarity helper function

GanShun:


JulienVdG:
I tested my nvram decoder above it and I can now use Attributes.ErasePolarity instead of 0xFF to properly read the image.
The code looks good to me.

PR #251, opened 2019-02-06
Nvram compact

JulienVdG:
This reimplements the perl script nvram-compact in fiano \o/

this goes on to of #250


JulienVdG:
thanks but it's missing tests...

JulienVdG:
one idea would be to place a fake rom (with only the nvram fv) in testdata and try the visitors on it
also maybe have json of the expected result.
what do you think ?

JulienVdG:
Done.

GanShun:
I do like the idea of a fake rom. I've been wanting to create a full fake rom though, with a fake ifd and all so we can test the IFD portion of the uefi package. We could also add the nvram in there, rather than just have an NVRAM only rom

JulienVdG:
I tend to prefer to test per feature, but I see your point.
What you need is just to play with extract / save and make your fake rom image starting from an existing rom and making sure you create your own files to replace the extracted content (or copy the minimal ones).
But it takes time ...

pkg/visitors/nvramcompact.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"github.com/linuxboot/fiano/pkg/guid"
+	"github.com/linuxboot/fiano/pkg/uefi"
+)
+
+func compactNVarStore(s *uefi.NVarStore) error {
+	var keepEntries []*uefi.NVar
+	linkedNVar := make(map[uint64]*uefi.NVar)
+	// Find Data entries and associated metadata entries
+	for _, v := range s.Entries {
```
GanShun: https://github.com/linuxboot/linuxboot/blob/master/bin/nvram-compact seems to imply that the name is stored on the first of the links to the nvar variable, while the last contains the actual value. Would we lose the name of the nvar in this case?

pkg/visitors/nvramcompact.go
```diff
@@ -0,0 +1,92 @@
...
+			return err
+		}
+		// call the compact function
+		return compactNVarStore(f)
+	}
+	return f.ApplyChildren(v)
+}
+
+func init() {
+	RegisterCLI("nvram-compact", "compact nvram content by removing old version of variables", 0, func(args []string) (uefi.Visitor, error) {
```
GanShun:
```suggestion
	RegisterCLI("nvram-compact", "compact nvram content by removing old versions of variables", 0, func(args []string) (uefi.Visitor, error) {
```

pkg/visitors/nvramcompact.go
JulienVdG: Yes the first has the name, GUID (plain or offset in the guidstore) and possible content.
The next links are updates to the value (like a journal).
During the parsing the name and GUID are copied in the NVar go struct (see: https://github.com/linuxboot/fiano/blob/master/pkg/uefi/nvram.go#L358 ) so we don't need to copy them in compact, only just select the last one.

Then we rebuild a `FullNVarEntry` that contains name, GUID and content.

pkg/visitors/find.go
```diff
@@ -224,6 +231,27 @@ func FindDXEFV(f uefi.Firmware) (*uefi.FirmwareVolume, error) {
 	return FindEnclosingFV(f, file)
 }
 
+// FindNVarPredicate is a generic predicate for searching NVar only.
+func FindNVarPredicate(r string) (func(f uefi.Firmware) bool, error) {
+	searchRE, err := regexp.Compile("^(" + r + ")$")
+	fmt.Printf("JVDG regex: %v\n", searchRE)
```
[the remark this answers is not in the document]
JulienVdG: I will drop them, thanks for pointing them

pkg/uefi/nvram.go
```diff
@@ -552,7 +552,7 @@ func (v *NVar) Assemble(content []byte, checkOnly bool) error {
 	//check/update Header.Size
 	if checkOnly {
 		if v.Header.Size != uint16(vData.Len()) {
-			return fmt.Errorf("NVAR size mismatch, expected %v got %v", v.DataOffset, vData.Len())
```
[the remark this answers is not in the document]
JulienVdG: Thanks, I found it during OSFC Hackathon when I did build the test fake rom.

pkg/visitors/nvramcompact.go
JulienVdG: done

pkg/visitors/find.go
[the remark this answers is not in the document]
JulienVdG: done

pkg/visitors/find.go
```diff
@@ -224,6 +231,27 @@ func FindDXEFV(f uefi.Firmware) (*uefi.FirmwareVolume, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	return func(f uefi.Firmware) bool {
+		switch f := f.(type) {
+		case *uefi.NVar:
+			//return searchRE.MatchString(f.Name)
+			ret := searchRE.MatchString(f.Name)
+			if ret {
+				fmt.Printf("JVDG matched: %v\n", f.Name)
```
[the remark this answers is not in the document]
JulienVdG: done

pkg/visitors/nvramcompact.go
GanShun: I see! ok

PR #253, opened 2019-02-11
remove futk

rminnich:
It's going to be in a separate repo so we can test building utk package
users out of tree.

That does not work currently. It should.

rminnich@xcpu:~/projects/linuxboot/ForthUTK$ go build
./ForthUTK.go:100:3: unknown field 'Pad' in struct literal of type visitors.Remove
./ForthUTK.go:200:4: unknown field 'Pad' in struct literal of type visitors.Remove
./ForthUTK.go:373:11: assignment mismatch: 2 variables but forth.Eval returns 1 values
./ForthUTK.go:473:5: r.Undo undefined (type *visitors.Remove has no field or method Undo)
./ForthUTK.go:532:4: f.Newop undefined (type forth.Forth has no field or method Newop)
./ForthUTK.go:537:9: assignment mismatch: 2 variables but forth.Eval returns 1 values
./ForthUTK.go:556:10: assignment mismatch: 2 variables but forth.Eval returns 1 values


hugelgupf:
I'm totally confused about the reason for this, and the separate repo thing. Can you explain?

JulienVdG:
Hello, If I understand correctly the idea is : 
 - UTK is untested as a package that can be used by other software
 - ForthUTK is not really part of UTK and should have it;s own repo anyway
 - separating ForthUTK allows testing UTK as an external package

:wink:

PR #255, opened 2019-02-13
Fix remove_dxes_except corner cases

JulienVdG:
The parseBlackList function create a regex with several string option,
FindFilePredicate encapsulate it to match whole word only but we have
to take care of the blacklist case where the regex string include '|'.

Signed-off-by: Julien Viard de Galbert <julien.viard-de-galbert@itrenew.com>

### DIFF
--- a/pkg/visitors/remove_test.go
+++ b/pkg/visitors/remove_test.go
@@ -112,3 +112,49 @@ func TestRemoveExcept(t *testing.T) {
 	}
 
 }
+
+func TestRemoveExceptBlackList(t *testing.T) {
+	// The following blacklist contains corner case:
+	// - start, mid or end of existing names but no full name
+	// - start, mid or end of existing GUID but no full GUID
+	// so it should behave as an empty blacklist, ie remove all
+	var blacklists = []string{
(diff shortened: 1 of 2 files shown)


pkg/visitors/remove_test.go
```diff
@@ -112,3 +112,54 @@ func TestRemoveExcept(t *testing.T) {
...
 
 }
+
+func TestRemoveExceptBlackList(t *testing.T) {
+	// The following blacklist contains corner case:
+	// - start, mid or end of existing names but no full name
+	// - start, mid or end of existing GUID but no full GUID
+	// so it should behave as an empty blacklist, ie remove all
+	var blacklists = []string{
+		`INEXISTING_FILENAME
```
GanShun: Could you use "" with \n here? or does that not work? That would make the formatting a little nicer :)

GanShun:
This looks good! just one nit

pkg/visitors/remove_test.go
JulienVdG: Right, at first I had a long list and no array... fixed

JulienVdG:
:angry: I clicked rebase and merge ... so why did github merge the commit instead of rebase it !!!

PR #263, opened 2019-04-02
Remove valid bit assertion.

GanShun:


GanShun:
Please check to see if this works for you!

JulienVdG:
Almost working, the State is always 0 after assemble, it need to be stored in json for it to work.
So removing the `json:"-"` from https://github.com/linuxboot/fiano/blob/master/pkg/uefi/file.go#L156 made it work.
Then integration test failed (as json changes) but this easy to fix.

GanShun:
Done, PTAL

PR #271, opened 2019-05-02
print warning but do not fail if NVRam can not be parsed

rminnich:
It's ok if we can't parse NVRam variables, just print a warning.


GanShun:
Does it work when you try to assemble it back out?

rminnich:
This fixes the problem with the sr630 FV

PR #274, opened 2019-06-07
pkg/fmap/fmap.go: Properly parse an fmap binary file

ArthurHeymans:
Instead of just looking for the __FMAP__ string parse the
whole binary until one with a fully valid header is found.

Signed-off-by: Arthur Heymans <arthur@aheymans.xyz>

### DIFF
--- a/pkg/fmap/fmap.go
+++ b/pkg/fmap/fmap.go
@@ -86,6 +86,22 @@ type Metadata struct {
 	Start uint64
 }
 
+func headerValid(h *Header) bool {
+	if h.VerMajor != 1 {
+		return false
+	}
+	// Check if some sensible value is used for the full flash size
+	if h.Size == 0 {
+		return false
+	}
+
+	// Name is specified to be null terminated single-word string without spaces
+	if bytes.Index(h.Name.Value[:], []byte("\x00")) == -1 {
+		return false
+	}
+	return true
+}
+
 // FlagNames returns human readable representation of the flags.
 func FlagNames(flags uint16) string {
 	names := []string{}
@@ -132,36 +148,54 @@ func Read(f io.Reader) (*FMap, *Metadata, error) {
 		return nil, nil, err
 	}
 
-	// Check for too many fmaps.
-	if bytes.Count(data, Signature) >= 2 {
-		return nil, nil, errMultipleFound
-	}
-
-	// Check for too few fmaps.
-	start := bytes.Index(data, Signature)
-	if start == -1 {
-		return nil, nil, errSigNotFound
-	}
-
-	// Reader anchored to the start of the fmap
-	r := bytes.NewReader(data[start:])
-
-	// Read fields.
+	// Loop over __FMAP__ occurrences until a valid header is found
+	start := 0
+	validFmaps := 0
 	var fmap FMap
-	if err := readField(r, &fmap.Header); err != nil {
-		return nil, nil, err
-	}
-	fmap.Areas = make([]Area, fmap.NAreas)
-	if err := readField(r, &fmap.Areas); err != nil {
-		return nil, nil, err
-	}
+	var fmapMetadata Metadata
+	for {
+		if start >= len(data) {
+			break
+		}
 
-	// Return useful metadata
-	fmapMetadata := Metadata{
-		Start: uint64(start),
-	}
+		next := bytes.Index(data[start:], Signature)
+		if next == -1 {
+			break
+		}
+		start += next
 
-	return &fmap, &fmapMetadata, nil
+		// Reader anchored to the start of the fmap
+		r := bytes.NewReader(data[start:])
+
+		// Read fields.
+		var testFmap FMap
+		if err := readField(r, &testFmap.Header); err != nil {
+			return nil, nil, err
+		}
+		if !headerValid(&testFmap.Header) {
+			start += len(Signature)
+			continue
+		}
+		fmap = testFmap
+		validFmaps++
+
+		fmap.Areas = make([]Area, fmap.NAreas)
+		err := readField(r, &fmap.Areas)
+		if err != nil {
+			return nil, nil, err
+		}
+		// Return useful metadata
+		fmapMetadata = Metadata{
+			Start: uint64(start),
+		}
+		start += len(Signature)
+	}
+	if validFmaps >= 2 {
+		return nil, nil, errMultipleFound
+	} else if validFmaps == 1 {
+		return &fmap, &fmapMetadata, nil
+	}
+	return nil, nil, errSigNotFound
 }
 
 // Write overwrites the fmap in the flash file.
--- a/pkg/fmap/fmap.go
+++ b/pkg/fmap/fmap.go
@@ -110,13 +110,18 @@ func FlagNames(flags uint16) string {
 	return strings.Join(names, "|")
 }
 
+var errEOF = errors.New("unexpected EOF while parsing fmap")
+
 func readField(r io.Reader, data interface{}) error {
 	// The endianness might depend on your machine or it might not.
 	if err := binary.Read(r, binary.LittleEndian, data); err != nil {
-		return errors.New("Unexpected EOF while parsing fmap")
+		return errEOF
 	}
 	return nil
 }
+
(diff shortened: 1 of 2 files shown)


pkg/fmap/fmap.go
```diff
@@ -127,36 +143,39 @@ func Read(f io.Reader) (*FMap, *Metadata, error) {
...
-		Start: uint64(start),
+		// Return useful metadata
+		fmapMetadata := Metadata{
+			Start: uint64(start),
+		}
+		return &fmap, &fmapMetadata, nil
 	}
-
-	return &fmap, &fmapMetadata, nil
+	return nil, nil, errors.New("Cannot find fmap signature")
```
rjoleary:
```suggestion
	return nil, nil, errors.New("Cannot find a valid fmap")
```

pkg/fmap/fmap.go
```diff
@@ -86,6 +86,22 @@ type Metadata struct {
 	Start uint64
 }
 
+// HeaderValid A basic check whether the fmap header is valid
+func HeaderValid(H *Header) bool {
+	const HeaderSize = 56
+	const AreaSize = 42
```
rjoleary: Instead of these two constants, can you use `binary.Size(&Header{})` and `binary.Size(&Area{})`?

pkg/fmap/fmap.go
```diff
@@ -86,6 +86,22 @@ type Metadata struct {
...
+
+	if H.VerMajor != 1 {
+		return false
+	}
+	if H.Size < uint32(HeaderSize+H.NAreas*AreaSize) {
```
rjoleary: Is there a reason this is `<` rather than `==`?

pkg/fmap/fmap.go
ArthurHeymans: ok thx

pkg/fmap/fmap.go
ArthurHeymans: So this is checking if the total flash size is smaller than the size of the fmap header struct + fmap entries struct. '< 'is correct since you generally want something else than an fmap in the flash, but this check might as well have been H.Size > 0...

pkg/fmap/fmap.go
```diff
@@ -127,36 +143,57 @@ func Read(f io.Reader) (*FMap, *Metadata, error) {
 		return nil, nil, err
 	}
 
-	// Check for too many fmaps.
-	if bytes.Count(data, Signature) >= 2 {
-		return nil, nil, errors.New("Found multiple signatures")
-	}
+	// Loop over __FMAP__ occurrences until a valid header is found
+	start := 0
+	ValidFmaps := 0
```
rjoleary: Make the local variable start with a lowercase -- that's the convention I usually see in Go (although it doesn't seem to be enforced by the linter):

```suggestion
	validFmaps := 0
```

pkg/fmap/fmap.go
```diff
@@ -127,36 +143,57 @@ func Read(f io.Reader) (*FMap, *Metadata, error) {
...
-	if start == -1 {
-		return nil, nil, errors.New("Cannot find fmap signature")
-	}
+		next := bytes.Index(data[start:], Signature)
+		if next == -1 {
+			break
+		}
+		start += next
+
+		if start >= len(data) {
```
rjoleary: Is this ever false?

pkg/fmap/fmap.go
```diff
@@ -127,36 +143,57 @@ func Read(f io.Reader) (*FMap, *Metadata, error) {
...
+			Start: uint64(start),
+		}
+		start += len(Signature)
 	}
-
-	// Return useful metadata
-	fmapMetadata := Metadata{
-		Start: uint64(start),
+	if ValidFmaps >= 2 {
+		return nil, nil, errors.New("Found multiple fmap")
```
rjoleary:
```suggestion
		return nil, nil, errors.New("Found multiple fmaps")
```

pkg/fmap/fmap.go
```diff
@@ -127,36 +143,57 @@ func Read(f io.Reader) (*FMap, *Metadata, error) {
...
-	}
-	fmap.Areas = make([]Area, fmap.NAreas)
-	if err := readField(r, &fmap.Areas); err != nil {
-		return nil, nil, err
+		// Read fields.
+		var TestFmap FMap
+		if err := readField(r, &TestFmap.Header); err != nil {
+			return nil, nil, err
+		}
+		if headerValid(&TestFmap.Header) {
```
rjoleary: I'd invert this if-statement, so:

```
if !headerValid(&TestFmap.Header) {
    start += len(Signature)
    continue
}
fmap = TestFmap
ValidFmaps++
``

Then, you don't need the else branch.

File: pkg/fmap/fmap.go [hunk 8]
```diff
@@ -127,36 +143,57 @@ func Read(f io.Reader) (*FMap, *Metadata, error) {
...
-	// Read fields.
-	var fmap FMap
-	if err := readField(r, &fmap.Header); err != nil {
-		return nil, nil, err
-	}
-	fmap.Areas = make([]Area, fmap.NAreas)
-	if err := readField(r, &fmap.Areas); err != nil {
-		return nil, nil, err
+		// Read fields.
+		var TestFmap FMap
```
rjoleary:
```suggestion
		var testFmap FMap
```

pkg/fmap/fmap.go
```diff
@@ -86,6 +86,22 @@ type Metadata struct {
 	Start uint64
 }
 
+func headerValid(H *Header) bool {
```
hugelgupf: nit: convention would be to use `h *Header` - lowercase variable names

pkg/fmap/fmap.go
hugelgupf: bytes.Index would have to be wrong in some way?

pkg/fmap/fmap.go
hugelgupf: This would be the why: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

pkg/fmap/fmap.go
hugelgupf: Additionally, do not capitalize: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pkg/fmap/fmap_test.go
```diff
@@ -120,7 +120,7 @@ func TestNoSignature(t *testing.T) {
 	fakeFlash := bytes.Repeat([]byte{0x53, 0x11, 0x34, 0x22}, 94387)
 	r := bytes.NewReader(fakeFlash)
 	_, _, err := Read(r)
-	expected := "Cannot find fmap signature"
```
hugelgupf: This is our fault, but you may as well fix it. Rather than comparing error strings, compare the error itself:

```go
var errSigNotFound = errors.New("cannot find FMAP signature")

func Read(...) ... {

  return errSigNotFound
}
```

and in the test
```go
_, _, got := Read(r)
if want := errSigNotFound; got != want {
  t.Errorf("Read(%v) = %v, want %v", r, got, want)
}
```

Please use `want` instead of `expected`, and place what you got before what you want as in the example. Don't do "expected: %s, got: %s". (We do it wrong below.)

Covered also by https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures

ArthurHeymans: Many thanks for the suggestions. It's my first contribution to a go project.

pkg/fmap/fmap.go
```diff
@@ -86,6 +86,22 @@ type Metadata struct {
 	Start uint64
 }
 
+func headerValid(h *Header) bool {
```
rjoleary: `Header` should be lowercase here.

ArthurHeymans: The struct is defined with an upper case. Should it be unexported?

rjoleary: Nevermind, you're right! I don't know what I was thinking.

rjoleary:
I'm going to merge this! Thanks for the patch!

Issue #275, opened 2019-07-08
Default to system XZ


GanShun:
We should default to using the system's XZ command, falling back to the Go one only if one can't be found. We've encountered many new people running out of space due to the Go xz compressor.

PR #276, opened 2019-07-09
Compression: default to system xz

pallaud:
Addresses issue #275: We should default to using the system's XZ command, falling back to the Go one only if one can't be found. We've encountered many new people running out of space due to the Go xz compressor.

### DIFF
--- a/pkg/compression/compression.go
+++ b/pkg/compression/compression.go
@@ -10,11 +10,12 @@ package compression
 
 import (
 	"flag"
+	"os/exec"
 
 	"github.com/linuxboot/fiano/pkg/guid"
 )
 
-var xzPath = flag.String("xzPath", "", "Path to system xz command used for lzma encoding. If unset, an internal lzma implementation is used.")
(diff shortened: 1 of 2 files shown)


pkg/compression/compression.go
```diff
@@ -9,13 +9,11 @@
 package compression
 
 import (
-	"flag"
+	"os/exec"
 
 	"github.com/linuxboot/fiano/pkg/guid"
 )
 
-var xzPath = flag.String("xzPath", "", "Path to system xz command used for lzma encoding. If unset, an internal lzma implementation is used.")
```
rjoleary: I think it would still be good to have this option, but have it default to "xz".

JulienVdG: +1

The rest looks good by the way ! Thanks for adding a test !

pallaud: Done! I kept the LookPath change, figured it wouldn't hurt :)

PR #278, opened 2019-07-13
Added FSPInfo tool

insomniacslk:
Only FSP 2.0 is supported at the moment. More to come.

Signed-off-by: Andrea Barberio <insomniac@slackware.it>


cmds/fspinfo/main.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/linuxboot/fiano/pkg/fsp"
+	"github.com/linuxboot/fiano/pkg/uefi"
+)
+
+func main() {
+	flag.Parse()
+
+	if flag.Arg(0) == "" {
+		panic("missing filename")
```
insomniacslk: maybe I should not use `panic` here

insomniacslk:
I've added the [Fsp.fd](https://github.com/IntelFsp/FSP/blob/master/ApolloLakeFspBinPkg/FspBin/Fsp.fd) binary from Intel's FSP repository. This should be OK since they're under the BSD license, but worth discussing.

cmds/fspinfo/main.go
hugelgupf: log.Fatalf

insomniacslk: I already replaced it with fmt.Fprintf and os.Exit... I am normally wary of log.Fatal because it doesn't honor deferred functions, but in this case it's not an issue. I'll change it to log.Fatal

insomniacslk:
I am no `dep` expert. I ran `dep ensure` which added the `uio` vendored dependency, but it seems an old one (`NewLittleEndianBuffer` is missing). If I run `dep ensure -update github.com/u-root/u-root` however it doesn't get updated. @hugelgupf do you know how can I fix that?

cmds/fspinfo/main.go
insomniacslk: done

insomniacslk:
this should be ready to review now

cmds/fspinfo/main.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		log.Fatalf("Error: cannot read input file: %v", err)
+	}
+
+	fv, err := uefi.NewFirmwareVolume(data, 0, false)
+	if err != nil {
+		log.Fatalf("Error: cannot parse Firmware Volume: %v", err)
+	}
+	if len(fv.Files) < 1 {
+		log.Fatal("Error: Firmware Volume ha no files")
```
osresearch: typo

pkg/fsp/header.go
```diff
@@ -0,0 +1,193 @@
...
+func (ca ComponentAttribute) String() string {
+	var attrs []string
+	if uint16(ca)&0x01 == 1 {
+		attrs = append(attrs, "ReleaseBuild")
+	} else {
+		attrs = append(attrs, "DebugBuild")
+	}
+	if uint16(ca)&0x03 == 1 {
+		attrs = append(attrs, "OfficialRelease")
+		attrs = append(attrs, "OfficialRelease")
```
osresearch: double?

pkg/fsp/header.go
```diff
@@ -0,0 +1,193 @@
...
+	if err := buf.FinError(); err != nil {
+		return nil, err
+	}
+	if !bytes.Equal(f.Signature[:], Signature[:]) {
+		return nil, fmt.Errorf("invalid signature %v; want %v", f.Signature, Signature)
+	}
+	if f.HeaderLength != CurrentHeaderLength {
+		return nil, fmt.Errorf("invalid header length %d; want %d", f.HeaderLength, CurrentHeaderLength)
+	}
+	if !bytes.Equal(f.Reserved1[:], []byte{0, 0}) {
```
osresearch: should all reserved bytes be checked?

pkg/fsp/header.go
```diff
@@ -0,0 +1,193 @@
...
+	f.Reserved3[2] = buf.Read8()
+	f.Reserved3[3] = buf.Read8()
+	f.NotifyPhaseEntryOffset = buf.Read32()
+	f.FSPMemoryInitEntryOffset = buf.Read32()
+	f.TempRAMExitEntryOffset = buf.Read32()
+	f.FSPSiliconInitEntryOffset = buf.Read32()
+	if err := buf.FinError(); err != nil {
+		return nil, err
+	}
+	if !bytes.Equal(f.Signature[:], Signature[:]) {
```
osresearch: does it make sense to move the signature and version checks to the top before the other fields are assigned?

osresearch:
> I've added the [Fsp.fd](https://github.com/IntelFsp/FSP/blob/master/ApolloLakeFspBinPkg/FspBin/Fsp.fd) binary from Intel's FSP repository. This should be OK since they're under the BSD license, but worth discussing.

They are not licensed under the BSD - there is a no reverse engineering clause:

> Redistribution and use in binary form, without modification, are permitted
provided that the following conditions are met:
> * Redistributions must reproduce the above copyright notice and the following disclaimer in the documentation and/or other materials provided with the distribution.
> * Neither the name of Intel Corporation nor the names of its suppliers may be used to endorse or promote products derived from this software without specific prior written permission.
> * No reverse engineering, decompilation, or disassembly of this software is permitted.

If the FD file is included in the tree it needs to include a copy of this license.

pkg/fsp/header.go
insomniacslk: I'd say yes - according to the [FSP specification](https://www.intel.com/content/dam/www/public/us/en/documents/technical-specifications/fsp-architecture-spec-v2.pdf), section 2.2.1:
```
All reserved fields defined in this specification must be zero unless stated otherwise.
```

pkg/fsp/header.go
insomniacslk: indeed, it would, I'll change that.

In the future I plan to also support the previous header revisions, which means first parsing the common header, and then assign to the appropriate implementation of that header revision.

insomniacslk:
> If the FD file is included in the tree it needs to include a copy of this license.

Good point! I've added the FSP_License.pdf in the test_blobs directory, along with a README that explains where these are coming from

pkg/fsp/header_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+import (
+	"bytes"
+	"io/ioutil"
+	"testing"
+)
+
+const (
+	// From https://github.com/IntelFsp/FSP/blob/master/ApolloLakeFspBinPkg/FspBin/Fsp.fd
+	// under the BSD license.
```
JulienVdG: As pointed by Trammel, it's not just a BSD license, also update this comment. ie:
`under the FSP License https://github.com/IntelFsp/FSP/blob/master/FSP_License.pdf.`

insomniacslk: whoops - I updated the content of `test_blobs` but not this comment

insomniacslk: fixed

insomniacslk:
this should be ready again for review

cmds/fspinfo/main.go
```diff
@@ -0,0 +1,75 @@
...
+		log.Fatalf("Error: cannot parse FSP Info Header: %v", err)
+	}
+	j, err := json.MarshalIndent(hdr, "", "    ")
+	if err != nil {
+		log.Fatalf("Error: cannot marshal JSON: %v", err)
+	}
+	if *flagJSON {
+		fmt.Println(string(j))
+	} else {
+		fmt.Printf("Signature                   : %s\n", hdr.Signature)
```
GanShun: If you need custom prints, could you make a stringer for the fsp header? rather than to do it here.

cmds/fspinfo/README.md
```diff
@@ -0,0 +1,47 @@
...
+`fspinfo` is a library and command line to parse and present [FSP](https://github.com/IntelFsp/FSP/blob/master/FSP_License.pdf)
+information.
+
+Intel FSP specification 2.0 can be found at
+https://www.intel.com/content/dam/www/public/us/en/documents/technical-specifications/fsp-architecture-spec-v2.pdf
+.
+
+## Example
+
+Grab an FSP file at https://github.com/IntelFsp/FSP if you don't have one already.
```
GanShun: WHat is the license on this, can we check this in and add integration tests?

pkg/fsp/header.go
```diff
@@ -0,0 +1,206 @@
...
+	CommonHeaderLength = 12
+	HeaderV3Length     = 72
+	// FSP 2.0
+	CurrentSpecVersion = SpecVersion(0x20)
+	HeaderV3Revision   = 3
+)
+
+// InfoHeaderRev3 represents the FSP_INFO_HEADER structure revision 3 (FSP
+// 2.0) as defined by Intel.
+type InfoHeaderRev3 struct {
```
GanShun: Would prefer a stringer on this struct rather than have the prints in the cmd

cmds/fspinfo/README.md
GanShun: Never mind Ignore me.

pkg/fsp/header.go
insomniacslk: will do! Thanks for the review :)

insomniacslk: one question before I do that: my understanding of the stringers in Go is that it's used to describe a type with a string. While I believe there's no recommendation about this, I don't remember multi-line stringers, rather short representations of the object.

The prints in cmds/fspinfo are a verbose, multi-line representation. Instead of `String()`, should we have instead another method (e.g. `Summary()`)?

GanShun: Honestly, I'm ok with either. I'll leave it up to you.

insomniacslk: ok, I'll go with `Summary` (we can change our mind later anyway)

insomniacslk: updated

pkg/fsp/header.go
```diff
@@ -0,0 +1,231 @@
...
+	return ret
+}
+
+// ComponentAttribute represents the component attribute.
+type ComponentAttribute uint16
+
+func (ca ComponentAttribute) String() string {
+	var attrs []string
+	if uint16(ca)&0x01 == 1 {
+		attrs = append(attrs, "ReleaseBuild")
```
GanShun: Note that you can split this up into different functions to get the attributes, if you ever want to kinda like how file.go does it. Might be worth putting a TODO here.

pkg/fsp/header.go
```diff
@@ -0,0 +1,231 @@
...
+		attrs = append(attrs, "TypeFSP-M")
+	case 3:
+		attrs = append(attrs, "TypeFSP-S")
+	case 8:
+		attrs = append(attrs, "TypeFSP-O")
+	default:
+		attrs = append(attrs, "TypeFSPReserved")
+	}
+	ret := fmt.Sprintf("%#04x %s", uint16(ca), strings.Join(attrs, "|"))
+	// bits 2:12 are reserved
```
GanShun: Wait if bits 2:12 are reserved, how does case 8 above work?

pkg/fsp/header.go
```diff
@@ -0,0 +1,231 @@
...
+		attrs = append(attrs, "TestRelease")
+	}
+	switch uint16(ca) >> 12 {
+	case 1:
+		attrs = append(attrs, "TypeFSP-T")
+	case 2:
+		attrs = append(attrs, "TypeFSP-M")
+	case 3:
+		attrs = append(attrs, "TypeFSP-S")
+	case 8:
```
GanShun: Is this supposed to be 4?

pkg/fsp/header.go
```diff
@@ -0,0 +1,231 @@
...
+	// parse the common header first
+	buf := uio.NewLittleEndianBuffer(b[:CommonHeaderLength])
+	f.Signature[0] = buf.Read8()
+	f.Signature[1] = buf.Read8()
+	f.Signature[2] = buf.Read8()
+	f.Signature[3] = buf.Read8()
+	f.HeaderLength = buf.Read32()
+	f.Reserved1[0] = buf.Read8()
+	f.Reserved1[1] = buf.Read8()
+	f.SpecVersion = SpecVersion(buf.Read8())
```
GanShun: The advantage of using binary.Read is that since SpecVersion is a uint8, you don't need to cast it.

pkg/fsp/header.go
```diff
@@ -0,0 +1,231 @@
...
+
+// FromBytes parses an FSP_INFO_HEADER from a byte buffer.
+func FromBytes(b []byte) (*InfoHeaderRev3, error) {
+	if len(b) < CommonHeaderLength {
+		return nil, fmt.Errorf("short FSP Info Header length %d; want at least %d", len(b), CommonHeaderLength)
+	}
+	var f InfoHeaderRev3
+	// parse the common header first
+	buf := uio.NewLittleEndianBuffer(b[:CommonHeaderLength])
+	f.Signature[0] = buf.Read8()
```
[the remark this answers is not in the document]
GanShun: I also don't think you need the uio package here.

pkg/fsp/header_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+func TestFromBytesHeaderV3(t *testing.T) {
+	buf, err := ioutil.ReadFile(FSPTestFile)
+	if err != nil {
+		t.Errorf("Error opening test file %s: %v", FSPTestFile, err)
+	}
+	// in the ApolloLake FSP, the FSP header starts at byte 148. This can be
+	// extracted by parsing it as uefi.FirmwareVolume, but it's not relevant for
+	// this test, so let's go with raw offsets.
+	hdr, err := FromBytes(buf[148 : 148+HeaderV3Length])
```
GanShun: One possibility to avoid this is to dump the exact binary we care about, and get rid of the wrapping headers.

pkg/fsp/header_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"io/ioutil"
+	"testing"
+)
+
+const (
+	// From https://github.com/IntelFsp/FSP/blob/master/ApolloLakeFspBinPkg/FspBin/Fsp.fd
+	// under the FSP license. See README.md under `test_blobs`.
+	FSPTestFile = "test_blobs/ApolloLakeFspBinPkg/Fsp.fd"
+)
+
```
GanShun: Would be nice to have some tests for the attribute parsing. I think this could be improved by moving the attribute parsing to separate functions and testing those. Another test that would be nice is just to try to read a badly formatted fsp binary and check for the error thrown

insomniacslk:
ping

pkg/fsp/header.go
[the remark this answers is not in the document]
insomniacslk: thanks. Parsing the common header is trivial, but parsing the rest less so. At the moment I have a `FixedInfoHeader`, and an `InfoHeaderV3`, where the latter embeds the former. In order to read the remaining fields, I'd have to read each one individually.

Alternatively I could have a `FixedInfoHeader` plus some `RestOfTheInfoHeaderV3` (fictional name :) ), plus an `InfoHeaderV3` that embeds both. In this case I could use `buffer.Read` on the second part of the header too. However I don't like this approach.

Another alternative is re-reading the buffer into the final structure (`InfoHeaderV3`) after the revision information has been extracted. While I don't like reading twice (even for a small amount of bytes), this is the solution that smells less.

Any advice?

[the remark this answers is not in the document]
insomniacslk: I committed the latter (re-reading the buffer) for now, let me know if that's OK

cmds/fspinfo/main.go
```diff
@@ -0,0 +1,56 @@
...
+	fv, err := uefi.NewFirmwareVolume(data, 0, false)
+	if err != nil {
+		log.Fatalf("Error: cannot parse Firmware Volume: %v", err)
+	}
+	if len(fv.Files) < 1 {
+		log.Fatal("Error: Firmware Volume has no files")
+	}
+	file := fv.Files[0]
+	// FIXME why does FSPH start at +4?
+	hdr, err := fsp.FromBytes(file.Buf()[file.DataOffset+4 : file.DataOffset+4+fsp.HeaderV3Length])
```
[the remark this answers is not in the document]
insomniacslk: I'm not sure what to do - once I have the `uefi.File` object, which has this 4-byte header, what should I parse it with?

pkg/fsp/header.go
insomniacslk: this is already an attribute-specific method if that's what you mean. Otherwise can you expand on it?

pkg/fsp/header.go
insomniacslk: my bad, it's bits 11:2, the 8 above is correct

see https://www.intel.com/content/dam/www/public/us/en/documents/technical-specifications/fsp-architecture-spec-v2.pdf section 5.1.1 table 1

pkg/fsp/header.go
insomniacslk: no, wrong comment on my side, see above

pkg/fsp/header_test.go
insomniacslk: I prefer to use the original FSP files from Intel, because at some point I want to introduce some integration tests. I agree with you that for unit tests the raw sequence of bytes would be more appropriate, and I can put that in the unit test file

pkg/fsp/header_test.go
insomniacslk: ah now I understand what you meant above about splitting attribute parsing.

Do you have something like this in mind?
```
func (ca ComponentAttribute) Parse() []string {
    // parser and return an array of strings, e.g. []string{"ReleaseBuild", "OfficialRelease"}
}
```

cmds/fspinfo/main.go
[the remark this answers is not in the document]
insomniacslk: (replied below)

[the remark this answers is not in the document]
GanShun: Does that uefi.File contain a uefi.Section? Is it a raw section?

pkg/fsp/header.go
GanShun: I mean it more like how fileAttr is handled in pkg/uefi/file.go. The attributes field is a one byte field, but there are separate functions for IsLarge etc...

pkg/fsp/header.go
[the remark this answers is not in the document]
GanShun: re-reading the buffer is ok for now

pkg/fsp/header.go
GanShun: This would make it easier to unit test specific attributes.

cmds/fspinfo/main.go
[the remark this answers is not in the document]
insomniacslk: doesn't seem to be a section, if I am doing this right. I'm using `uefi.NewSection`, and I get the error

```
section size mismatch! Section has size 2572478, but buffer is 140 bytes big
```

However I couldn't figure out what `fileOrder` is, so I used 0 (which may be just plain wrong)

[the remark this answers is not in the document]
insomniacslk: My mistake. This works. Still unsure about file order though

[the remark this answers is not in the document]
insomniacslk: also, yes, it's a raw file, so sections are not parsed

pkg/fsp/header_test.go
insomniacslk: maybe I can test the CLI with the whole FSP file as an integration test, and the pkg with the golden header as a unit test. Would that work?

pkg/fsp/header.go
insomniacslk: oh I see - basically just attaching methods that test the flag bits

insomniacslk: done

cmds/fspinfo/main.go
```diff
@@ -0,0 +1,56 @@
...
+	flag.Parse()
+	if flag.Arg(0) == "" {
+		log.Fatal("Error: missing file name")
+	}
+	data, err := ioutil.ReadFile(flag.Arg(0))
+	if err != nil {
+		log.Fatalf("Error: cannot read input file: %v", err)
+	}
+
+	fv, err := uefi.NewFirmwareVolume(data, 0, false)
```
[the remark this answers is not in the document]
insomniacslk: done

pkg/fsp/header_test.go
insomniacslk: done that way

insomniacslk:
still TODO: the CLI only parses the first FV, so it only shows the first FSP component, I'll parse the other FVs too

cmds/fspinfo/main.go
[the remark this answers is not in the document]
GanShun: This is so stupid. Why did they use a raw file type? This way when you're parsing you have no idea whether to treat raw file contents as sections or not, because there's no signature and it could be garbage that you're reading as the size and type.

GanShun:
Sorry for the long delay, can you rebase and resolve the conflicts? I'm ok with this going in as is for now.

insomniacslk:
sure!

insomniacslk:
rebased, cc @GanShun

PR #280, opened 2019-07-22
New layout-table visitor to show base address and size of each element

JulienVdG:
Note: This was implemented as a copy of table, the main visitor logic is the same so it could also be implemented as options on the Table visitor.

Tell me if you want me to refactor it in the Table visitor, or if you prefer it like this as a separate code.

### DIFF
--- a/pkg/visitors/table.go
+++ b/pkg/visitors/table.go
@@ -21,6 +21,7 @@ type Table struct {
 	indent    int
 	offset    uint64
 	curOffset uint64
+	printRow  func(v *Table, node, name, typez interface{}, offset, length uint64)
 }
 
 // Run wraps Visit and performs some setup and teardown tasks.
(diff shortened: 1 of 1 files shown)


JulienVdG:
1. Ok will do.
2. I was not aware of that, thanks I'll update the depth logic in consequence.

JulienVdG:
Done!
I also tested with the OVMF rom image that has only the fake `BIOSregion` and fixed a few more corner cases.

pkg/visitors/table.go
```diff
@@ -26,57 +31,105 @@ func (v *Table) Run(f uefi.Firmware) error {
...
+		return v.printRow(f, fmt.Sprintf("%T", f), "", "", 0, 0)
 	}
 }
 
 func indent(n int) string {
 	return strings.Repeat(" ", n)
 }
 
-func (v *Table) printRow(f uefi.Firmware, node, name, typez interface{}) error {
+func (v *Table) printRow(f uefi.Firmware, node, name, typez interface{}, offset, dataOffset uint64) error {
```
GanShun: I highly recommend a refactor here. A large part of this code is dependent on if v.Layout vs otherwise.

We can adapt this for different table styles by making two printRow functions, and keeping a variable inside the table visitor struct 

type Table struct {
	W      *tabwriter.Writer
	indent int
	Layout    bool
	TopLevel  bool
	indent    int
	offset    uint64
	curOffset uint64
	depth     int
	print       func(f uefi.Firmware, node, name, typez interface{}, offset, dataOffset uint64) error 
}

and assigning the print based on whether Layout is set.

For even more code reuse, we can split the print function into two segments instead, one which does all the printing prior to applychildren called printData, and one which does all the printing after applychildren called printFooter, and this printRow would call v.printData(...) then f.ApplyChildren then v.printFooter(...)

GanShun: I think this would make it much clearer to the reader.

pkg/visitors/table.go
```diff
@@ -26,57 +31,105 @@ func (v *Table) Run(f uefi.Firmware) error {
...
+		if v.Layout {
+			fmt.Fprintf(v.W, "%s%v\t%v\t%#08x\t%#08x\n", indent(v2.indent), "Free", "", offset+f.FreeSpaceOffset, f.GUIDStoreOffset-f.FreeSpaceOffset)
+			fmt.Fprintf(v.W, "%s%v\t%v\t%#08x\t%#08x\n", indent(v2.indent), "GUIDStore", fmt.Sprintf("%d GUID", len(f.GUIDStore)), offset+f.GUIDStoreOffset, f.Length-f.GUIDStoreOffset)
+		} else {
+			fmt.Fprintf(v.W, "%s%v\t%v\t%v\t%#8x\n", indent(v2.indent), "Free", "", "", f.GUIDStoreOffset-f.FreeSpaceOffset)
+			fmt.Fprintf(v.W, "%s%v\t%v\t%v\t%#8x\n", indent(v2.indent), "GUIDStore", "", fmt.Sprintf("%d GUID", len(f.GUIDStore)), f.Length-f.GUIDStoreOffset)
+		}
+	case *uefi.File:
+		// Align
+		// TODO: do we need the complex align logic from assemble?
```
GanShun: You don't need it as long as the user has assembled the image prior, or has not made any modifications to the tree. I haven't forced an assemble unconditionally because that makes it slow though, but there runs the risk that if a user doesn't assemble after removing, table may display old/incorrect information.

pkg/visitors/table.go
```diff
@@ -26,57 +31,105 @@ func (v *Table) Run(f uefi.Firmware) error {
 
 // Visit applies the Table visitor to any Firmware type.
 func (v *Table) Visit(f uefi.Firmware) error {
+	var offset uint64
 	switch f := f.(type) {
 	case *uefi.FlashImage:
-		return v.printRow(f, "Image", "", "")
+		v.depth = v.indent + 1
```
GanShun: depth should not be = v.indent + 1. If we ever want to add more cli's that go one or two levels deeper, this would cause it to always only go one level deep, it should be depth++ in uefi.FlashImage (because we want to not count the flashimage), and not anywhere else. Then in 

RegisterCLI("layout-table", "print out offset and size information of top level firmware volumes in a pretty table", 0, func(args []string) (uefi.Visitor, error) {
		return &Table{Layout: true, TopLevel: true, depth: 1}, nil
	})

This way we can easily add more cli's that traverse different depths I think

pkg/visitors/table.go
JulienVdG: I did a first refactoring.
I'm bothered with the function in the struct... It can't use the receiver anymore, right ? So I'd have to give `v *Table` as an argument

pkg/visitors/table.go
JulienVdG: Ok, I'll drop the TODO then thanks

pkg/visitors/table.go
JulienVdG: I first though of using `Depth` in the interface ... and found it confusing:
If you ask for a depth of 2, you'll end with a depth of 4 on an Intel Image (2 + Image + IFD before BIOS Region)

So I opted for the `TopLevel` flag to be in the interface and the `depth` to only be an implementation internal... (but being in the same package makes the Exported/private less obvious).

I'll try your approach and see how it goes ;)

JulienVdG: Thanks, that works, I was probably missing the point that IFD is not a container, only uefi.FlashImage is...

pkg/visitors/table.go
JulienVdG: done

PR #284, opened 2019-09-06
Fix region parsing for weird bioses

GanShun:


rminnich:
could there be a bit more info in the commit message?

PR #285, opened 2019-10-14
ME partitions

JulienVdG:
ME Partition parsing, the goal is to spot a padding in the ME Region after the ME partitions so that this region can be shrunk in the IFD.

ME partition informations from http://me.bios.io/ME_blob_format

### DIFF
--- a/pkg/uefi/meregion.go
+++ b/pkg/uefi/meregion.go
@@ -4,8 +4,161 @@
 
 package uefi
 
+import (
+	"bytes"
+	"encoding/binary"
+	"encoding/hex"
+	"fmt"
+	"log"
+)
+
+// ME Partition parsing, the goal is to spot a padding in the ME Region
+// after the ME partitions so that this region can be shrunk in the IFD.
+//
+// ME partition informations from http://me.bios.io/ME_blob_format
+
+// MEFTPSignature is the sequence of bytes that an ME Flash Partition
+// Table is expected to start with.
+var (
+	MEFTPSignature = []byte{0x24, 0x46, 0x50, 0x54}
+)
+
+const (
+	// MEFTPSignatureLength represents the size of the ME FTP signature
+	MEFTPSignatureLength = 4
+	// MEPartitionTableEntryLength is the size of a partition table entry
+	MEPartitionTableEntryLength = 32
+)
+
+// MEFPT is the main structure that represents an ME Flash Partition Table.
+type MEFPT struct {
+	// Holds the raw buffer
+	buf []byte
+
+	PartitionCount    uint32
+	PartitionMapStart int
+	Entries           []MePartitionEntry
+	//Metadata for extraction and recovery
+	ExtractPath string
+}
+
+// MePartitionEntry is an entry in FTP
+type MePartitionEntry struct {
+	Name     MeName
+	Owner    [4]byte
+	Offset   uint32
+	Length   uint32
+	Reserved [4]uint32
+}
+
+// MeName represent 4 bytes with JSON string support
+// OK this is probably overkill!
+type MeName [4]byte
+
+// MarshalText converts MeName to a byte range (for JSON)
+func (n MeName) MarshalText() ([]byte, error) {
+	e := len(n)
+	for e > 0 && n[e-1] == 0 {
+		e--
+	}
+	return n[:e], nil
+}
+
+// UnmarshalText converts a byte range to MeName (for JSON)
+func (n *MeName) UnmarshalText(b []byte) error {
+	for i := range n[:] {
+		var v byte
+		if i < len(b) {
+			v = b[i]
+		}
+		n[i] = v
+	}
+	return nil
+}
+
+func (n MeName) String() string {
+	b, _ := n.MarshalText()
+	return string(b)
+}
+
+// FindMESignature searches for an Intel ME FTP signature
+func FindMESignature(buf []byte) (int, error) {
+	if bytes.Equal(buf[16:16+MEFTPSignatureLength], MEFTPSignature) {
+		// 16 + 4 since the descriptor starts after the signature
+		return 20, nil
+	}
+	if bytes.Equal(buf[:MEFTPSignatureLength], MEFTPSignature) {
+		// + 4 since the descriptor starts after the signature
+		return MEFTPSignatureLength, nil
+	}
+	return -1, fmt.Errorf("ME Flash Partition Table signature not found: first 20 bytes are:\n%s",
+		hex.Dump(buf[:20]))
+}
+
+// Buf returns the buffer.
+// Used mostly for things interacting with the Firmware interface.
+func (fp *MEFPT) Buf() []byte {
+	return fp.buf
+}
+
+// SetBuf sets the buffer.
+// Used mostly for things interacting with the Firmware interface.
+func (fp *MEFPT) SetBuf(buf []byte) {
+	fp.buf = buf
+}
+
+// Apply calls the visitor on the MEFPT.
+func (fp *MEFPT) Apply(v Visitor) error {
+	return v.Visit(fp)
+}
+
+// ApplyChildren calls the visitor on each child node of MEFPT.
+func (fp *MEFPT) ApplyChildren(v Visitor) error {
+	return nil
+}
+
+// NewMEFPT tries to create a MEFPT
+func NewMEFPT(buf []byte) (*MEFPT, error) {
+	o, err := FindMESignature(buf)
+	if err != nil {
+		return nil, err
+	}
+	if len(buf) < o+28 {
+		return nil, fmt.Errorf("ME section (%#x) too small for ME Flash Partition Table (%#x)", len(buf), o+28)
+	}
+	fp := &MEFPT{PartitionMapStart: o + 28}
+	r := bytes.NewReader(buf[o:])
+	if err := binary.Read(r, binary.LittleEndian, &fp.PartitionCount); err != nil {
+		return nil, err
+	}
+	l := fp.PartitionMapStart + MEPartitionTableEntryLength*int(fp.PartitionCount)
+	if len(buf) < l {
+		return nil, fmt.Errorf("ME section (%#x) too small for %d entries in ME Flash Partition Table (%#x)", len(buf), fp.PartitionCount, l)
+	}
+
+	fp.buf = make([]byte, l)
+	copy(fp.buf, buf[:l])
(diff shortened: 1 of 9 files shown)


pkg/uefi/meregion.go
```diff
@@ -0,0 +1,238 @@
+// Copyright 2018 the LinuxBoot Authors. All rights reserved
```
rjoleary: 2019

pkg/uefi/meregion.go
```diff
@@ -0,0 +1,238 @@
...
+	"encoding/binary"
+	"encoding/hex"
+	"fmt"
+	"log"
+)
+
+// ME Partition parsing, the goal is to spot a padding in the ME Region
+// after the ME partitions so that this region can be shrunk in the IFD.
+//
+// ME partition informations from http://me.bios.io/ME_blob_format
```
rjoleary: Place this comment directly in front of "package uefi". This will give the nicest output from running "go doc".

pkg/uefi/meregion.go
```diff
@@ -0,0 +1,238 @@
...
+
+	PartitionCount    uint32
+	PartitionMapStart int
+	Entries           []MePartitionEntry
+	//Metadata for extraction and recovery
+	ExtractPath string
+}
+
+// MePartitionEntry is an entry in FTP
+type MePartitionEntry struct {
```
rjoleary: Perhaps, for consistency with the other names, this should be called `MEPartitionEntry` (capitalizing the E).

pkg/uefi/meregion.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// MeName represent 4 bytes with JSON string support
+// OK this is probably overkill!
+type MeName [4]byte
+
+// MarshalText converts MeName to a byte range (for JSON)
+func (n MeName) MarshalText() ([]byte, error) {
+	e := len(n)
+	for e > 0 && n[e-1] == 0 {
```
rjoleary: I think this can be simplified to something like `return bytes.TrimRight(n[:], "\x00"), nil`

pkg/uefi/meregion.go
```diff
@@ -0,0 +1,238 @@
...
+// UnmarshalText converts a byte range to MeName (for JSON)
+func (n *MeName) UnmarshalText(b []byte) error {
+	for i := range n[:] {
+		var v byte
+		if i < len(b) {
+			v = b[i]
+		}
+		n[i] = v
+	}
+	return nil
```
rjoleary: Perhaps, return an error if `len(b) > len(MeName)`. The reasoning is you don't want to silently truncate the name.

pkg/uefi/meregion.go
```diff
@@ -0,0 +1,238 @@
...
+	e := len(n)
+	for e > 0 && n[e-1] == 0 {
+		e--
+	}
+	return n[:e], nil
+}
+
+// UnmarshalText converts a byte range to MeName (for JSON)
+func (n *MeName) UnmarshalText(b []byte) error {
+	for i := range n[:] {
```
rjoleary: It should be possible to use Go's builtin `copy` function to copy from `b` to `n`. Something like `copy((*n)[:], b)`. The copy function uses the minimum size of both slices, so this should be safe whatever the size are.

pkg/uefi/meregion.go
```diff
@@ -0,0 +1,238 @@
...
+			v = b[i]
+		}
+		n[i] = v
+	}
+	return nil
+}
+
+func (n MeName) String() string {
+	b, _ := n.MarshalText()
+	return string(b)
```
rjoleary: You might want to do a `return fmt.Sprintf("%q", string(b))` just in case the string contains special characters.

pkg/uefi/meregion.go
```diff
@@ -0,0 +1,238 @@
...
+func (rr *MERegion) Apply(v Visitor) error {
+	return v.Visit(rr)
+}
+
+// ApplyChildren calls the visitor on each child node of MERegion.
+func (rr *MERegion) ApplyChildren(v Visitor) error {
+	if rr.FPT == nil {
+		return nil
+	}
+	if err := rr.FPT.Apply(v); err != nil {
```
rjoleary: This can be simplified to "return rr.FPT.Apply(v)". You don't need the if-statement.

pkg/uefi/meregion.go
JulienVdG: thanks !

pkg/uefi/meregion.go
JulienVdG: right thanks, I rarely use go doc... I should check it more

pkg/uefi/meregion.go
JulienVdG: Right, also more consistent with go conventions for initials in names

pkg/uefi/meregion.go
JulienVdG: Right, thanks

pkg/uefi/meregion.go
JulienVdG: Oh, nice, I didn't though of using a Trim function, perfect

pkg/uefi/meregion.go
JulienVdG: I see but the copy will not fill the remaining bytes with 0 when b is shorter than 4.
The Zero value should be ok, but can UnmarshalText be called on a reused MEName ?
... well I'll add some test and check that.

pkg/uefi/meregion.go
JulienVdG: Not sure about that one, %q will add an extra quoting I don't think we want on the table output

pkg/uefi/meregion.go
JulienVdG: except `golint` disagrees:  pkg/uefi/meregion.go:5:1: package comment should be of the form "Package uefi ..."

rjoleary: Oh right, I forgot all of uefi is in one package.

pkg/uefi/meregion.go
```diff
@@ -0,0 +1,228 @@
...
+
+// MEFTPSignature is the sequence of bytes that an ME Flash Partition
+// Table is expected to start with.
+var (
+	MEFTPSignature = []byte{0x24, 0x46, 0x50, 0x54}
+)
+
+const (
+	// MEFTPSignatureLength represents the size of the ME FTP signature
+	MEFTPSignatureLength = 4
```
rjoleary: Instead of having this constant, can you use `len(MEFTPSignature)` throughout?

pkg/uefi/meregion.go
```diff
@@ -0,0 +1,228 @@
...
+	return nil
+}
+
+func (n MEName) String() string {
+	b, _ := n.MarshalText()
+	return string(b)
+}
+
+// FindMESignature searches for an Intel ME FTP signature
+func FindMESignature(buf []byte) (int, error) {
```
rjoleary: It looks like this function returns the start of the descriptor. Perhaps it would be better to name it `FindMEDescriptor`.

pkg/uefi/meregion.go
```diff
@@ -0,0 +1,228 @@
...
+	if bytes.Equal(buf[16:16+MEFTPSignatureLength], MEFTPSignature) {
+		// 16 + 4 since the descriptor starts after the signature
+		return 20, nil
```
rjoleary: Return `16 + len(MEFTPSignature)` to be a bit more clear.

pkg/uefi/meregion.go
```diff
@@ -0,0 +1,228 @@
...
+
+// ME Partition parsing, the goal is to spot a padding in the ME Region
+// after the ME partitions so that this region can be shrunk in the IFD.
+//
+// ME partition informations from http://me.bios.io/ME_blob_format
+
+// MEFTPSignature is the sequence of bytes that an ME Flash Partition
+// Table is expected to start with.
+var (
+	MEFTPSignature = []byte{0x24, 0x46, 0x50, 0x54}
```
rjoleary: Add a comment here saying this is the string "$FPT"

pkg/uefi/meregion.go
```diff
@@ -0,0 +1,228 @@
...
+	return nil
+}
+
+// NewMEFPT tries to create a MEFPT
+func NewMEFPT(buf []byte) (*MEFPT, error) {
+	o, err := FindMESignature(buf)
+	if err != nil {
+		return nil, err
+	}
+	if len(buf) < o+28 {
```
rjoleary: Make the `28` a constant.

pkg/uefi/meregion.go
JulienVdG: I endedup creating a new MEName, using `copy` and then affecting it to the receiver.
It might be too much care, but MEName is only 4 bytes anyway so it does not cost much !

pkg/uefi/meregion.go
```diff
@@ -0,0 +1,228 @@
...
+	copy(fp.buf, buf[:l])
+	if err := fp.parsePartitions(); err != nil {
+		return nil, err
+	}
+	return fp, nil
+}
+
+func (fp *MEFPT) parsePartitions() error {
+	r := bytes.NewReader(fp.buf[fp.PartitionMapStart:])
+	for i := 0; i < int(fp.PartitionCount); i++ {
```
rjoleary: I'm pretty sure you don't need the for loop. You can do something like this: https://play.golang.org/p/UXKLQ89-zZr

pkg/uefi/meregion.go
JulienVdG: Right but then `FlashSignatureLength` in flash.go should also be changed ;)

pkg/uefi/meregion.go
JulienVdG: Done

pkg/uefi/meregion.go
JulienVdG: Done

pkg/uefi/meregion.go
JulienVdG: Done

pkg/uefi/meregion.go
JulienVdG: Wow ! So cool !

I'm still doing go thinking too much like C :wink:

JulienVdG:
Thanks a lot for the review !
I learned a few more tricks to avoid loops.
I hope I addressed all your concerns.

Issue #273, opened 2019-05-27
 Firmware volume creation/replacement


JulienVdG:
The linuxboot perl implementation has a create-fv for that, I'm not sure it's necessary here, all depends on how utk can handle the firmware volumes.

I think the insert visitor should be enough to fill existing volumes with extra files.

But there is nothing currently to add a fv.
The use case is for existing uefi platform where changing the flash layout is not possible but the flash includes some empty area. Creating a fv in each of those area will allow us to store extra data.

utk already find the empty area and creates `BIOSPadding` but there does not seam to be a way to touch them currently.

JulienVdG:
@rjoleary We did talk a little about this during osf hackathon. I remember the idea of partition editing...
Have you given more thought to it ? Do you have some suggestion for the implementation ?

GanShun:
How much customization do you want to allow for FV creation? One (terrible and tedious) way I suspect you can do it right now is to extract to an entirely new directory, add a new FV in and modify the json, then reassemble the image. However we can do better. 

1) One simple way is to just add a visitor that inserts an FV into an empty region given a starting offset, and checks if there are other firmware volumes in that spot. The problem is that it's not clear to the user what offsets are safe and what aren't. We could add something that dumps the top level layout so we can see free space clearly. I assume you're trying to work with an Intel image. Is the free space you're talking about part of the bios region? are you only trying to replace the BIOSPadding? or are you trying to use other empty areas such as the ME region or PD region? 

2) What happens if the user wants to insert a nested FV? I don't currently see use cases for that, so we could just not allow it or ignore it. (Note that a user could still insert nested FVs today if they wrap the FV in a File and insert that)

3) More generally I want to also be able to extend existing FVs into empty space after them. I'm not sure what the best way to do this would be. One way is to dump a high level summary that only has detail down to the top level FVs. Kind of like the way ifdtool dumps a small summary with offsets into a file, and you can modify that and have the tool do the modification. This seems like a weird exception though and doesn't really fit into our current model.

JulienVdG:
Thanks for the answers.
Yes extract, edit, reassemble works and I've done a few tests with this approach but it's not a solution for a production tool.

1. Your point 1 should be easy to implement (as well as a summary of the top level layout as a different version of the table visitor).
[...]

PR #286, opened 2019-10-15
Create fv

JulienVdG:
This is related to https://github.com/linuxboot/fiano/issues/273 option 1.

~~It's a work in progress with several hardcodded values, but it starts to work so initial review is welcome !~~

As per Ron's comment bellow, I think this can be reviewed, "TODO included" :wink:

### DIFF
--- a/pkg/visitors/createfv.go
+++ b/pkg/visitors/createfv.go
@@ -78,120 +78,37 @@ func (v *CreateFV) Visit(f uefi.Firmware) error {
 			return fmt.Errorf("Cannot create FV at %#x (+%#x) no matching BIOS Pad found", v.AbsOffset, v.Size)
 		}
 		v.found = true
-		return v.createFV(f, offset, bp, idx)
+		fv, err := createEmptyFirmwareVolume(v.AbsOffset-offset, v.Size, &v.Name)
+		if err != nil {
+			return err
+		}
+		return insertFVinBP(f, v.AbsOffset-offset, bp, idx, fv)
 	}
 
 	return f.ApplyChildren(v)
 }
 
-func (v *CreateFV) createFV(br *uefi.BIOSRegion, offset uint64, bp *uefi.BIOSPadding, idx int) error {
+func insertFVinBP(br *uefi.BIOSRegion, offset uint64, bp *uefi.BIOSPadding, idx int, fv *uefi.FirmwareVolume) error {
 	// Copy the Elements before the modified BIOS Pad
 	newElements := make([]*uefi.TypedFirmware, idx, len(br.Elements))
 	copy(newElements, br.Elements[:idx])
 
 	bpBuf := bp.Buf()
 	// keep part of the BIOS Pad before the new fv if needed
-	if offset+bp.Offset < v.AbsOffset {
-		hbp, err := uefi.NewBIOSPadding(bpBuf[:v.AbsOffset-offset-bp.Offset], bp.Offset)
+	if bp.Offset < offset {
+		hbp, err := uefi.NewBIOSPadding(bpBuf[:offset-bp.Offset], bp.Offset)
 		if err != nil {
 			return err
 		}
 		newElements = append(newElements, uefi.MakeTyped(hbp))
 	}
 
-	// TODO: can this be refactored with the code in repack.go and assemble.go ?
-	fv := &uefi.FirmwareVolume{} // new Firmware Volume
-	// Set up volume header first.
-	fv.FileSystemGUID = *uefi.FFS2
-	fv.Signature = binary.LittleEndian.Uint32([]byte("_FVH"))
-	// TODO: retrieve all details from (all) other fv in BIOS Region
-	fv.Attributes = 0x0004FEFF
-	fv.Revision = 2
-	// Create Blocks
-	fv.Blocks = make([]uefi.Block, 2)
-	fv.Blocks[0] = uefi.Block{Size: 4096, Count: uint32(v.Size / 4096)}
-	fv.Blocks[1] = uefi.Block{}
-	// Calculate the HeaderLen field
-	fv.HeaderLen = uint16(uefi.FirmwareVolumeFixedHeaderSize + int(unsafe.Sizeof(uefi.Block{}))*len(fv.Blocks))
-
-	fv.DataOffset = uint64(fv.HeaderLen) // unless we add the extended header
-	fv.Length = v.Size
-
-	// TODO: should the extended header offset be computed ?
-	fv.ExtHeaderOffset = 0x60
-	fv.FVName = v.Name
-	fv.ExtHeaderSize = uefi.FirmwareVolumeExtHeaderMinSize
-
-	// Generate binary header.
-	header := new(bytes.Buffer)
-	err := binary.Write(header, binary.LittleEndian, fv.FirmwareVolumeFixedHeader)
-	if err != nil {
-		return fmt.Errorf("unable to construct binary header of new firmware volume: got %v", err)
-	}
-	for _, b := range fv.Blocks {
-		err = binary.Write(header, binary.LittleEndian, b)
-		if err != nil {
-			return fmt.Errorf("unable to construct binary header of new firmware volume: got %v", err)
-		}
-
-	}
-	buf := header.Bytes()
-
-	// Checksum the header
-	sum, err := uefi.Checksum16(buf[:fv.HeaderLen])
-	if err != nil {
-		return err
-	}
-	newSum := 0 - sum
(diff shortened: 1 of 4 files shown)


pkg/visitors/createfv.go
```diff
@@ -0,0 +1,192 @@
...
+		return nil
+	}
+	var offset, end uint64
+	switch f := f.(type) {
+	case *uefi.BIOSRegion:
+		if f.FRegion != nil {
+			offset = uint64(f.FRegion.BaseOffset())
+			end = uint64(f.FRegion.EndOffset())
+		}
+		fmt.Printf("BIOS     %#08x-%#08x\n", offset, end)
```
rminnich: I assume at some point this becomes a debug print?

pkg/visitors/createfv.go
```diff
@@ -0,0 +1,192 @@
...
+	// keep part of the BIOS Pad after the new fv if needed
+	if v.AbsOffset-offset-bp.Offset+v.Size < uint64(len(bpBuf)) {
+		tbp, err := uefi.NewBIOSPadding(bpBuf[v.AbsOffset-offset-bp.Offset+v.Size:], v.AbsOffset-offset+v.Size)
+		if err != nil {
+			return err
+		}
+		newElements = append(newElements, uefi.MakeTyped(tbp))
+	}
+
+	// Keep the remaining
```
rminnich: remaining what?

pkg/visitors/createfv.go
JulienVdG: I'm going to drop it entierly

pkg/visitors/createfv.go
JulienVdG: oops, thanks!

JulienVdG:
OK, I'll check that the modified system can boot and will upstream an initial version. Thanks !

JulienVdG:
This version did generate a valid image for winterfell !

JulienVdG:
And I just checked on Leopard (OCP V1 with DDR3) and it can now start linuxboot !!!
It needed the Volume GUID in the extended header to detect the new volume which is not supported by the perl version of create-fv.

JulienVdG:
There are still some hardcoded value and refactoring to do but I think it can wait a later PR.

pkg/visitors/createfv_test.go
```diff
@@ -0,0 +1,160 @@
...
+		var offset, length uint64
+		switch f := e.Value.(type) {
+		case *uefi.FirmwareVolume:
+			offset = f.FVOffset
+			length = f.Length
+		case *uefi.BIOSPadding:
+			offset = f.Offset
+			length = uint64(len(f.Buf()))
+		default:
+			t.Fatalf("Unextpected Element at %d: %s", i, e.Type)
```
JulienVdG: typo
```suggestion
			t.Fatalf("Unexpected Element at %d: %s", i, e.Type)
```

pkg/visitors/createfv.go
```diff
@@ -0,0 +1,240 @@
...
+	if err != nil {
+		return err
+	}
+	if !v.found {
+		return fmt.Errorf("Cannot create FV at %#x (+%#x) no BIOS region found", v.AbsOffset, v.Size)
+	}
+	return nil
+}
+
+// Visit applies the Remove visitor to any Firmware type.
```
GanShun: Remove -> CreateFV

pkg/visitors/createfv.go
```diff
@@ -0,0 +1,240 @@
...
+		idx := -1
+		var bp *uefi.BIOSPadding
+	L:
+		for i, e := range f.Elements {
+			switch f := e.Value.(type) {
+			case *uefi.BIOSPadding:
+				bp = f
+				bpOffset := offset + f.Offset
+				bpEnd := bpOffset + uint64(len(f.Buf()))
+				if v.AbsOffset >= bpOffset && v.AbsOffset+v.Size <= bpEnd {
```
GanShun: This won't work if for whatever reason we have two biospads that are contiguous (Not sure why that'd be the case though, but it might happen if we delete FVs in the future.)

GanShun: I'm not happy about the way I did biospads in the past, maybe it's time to revisit them and remove biospads. Alternatively we should do something that joins contiguous biospad regions so that this won't happen.

pkg/visitors/createfv.go
```diff
@@ -0,0 +1,240 @@
...
+			return err
+		}
+		newElements = append(newElements, uefi.MakeTyped(hbp))
+	}
+
+	// Add the FV
+	newElements = append(newElements, uefi.MakeTyped(fv))
+
+	// keep part of the BIOS Pad after the new fv if needed
+	if offset-bp.Offset+fv.Length < uint64(len(bpBuf)) {
```
GanShun: you can assign offset-bp.Offset+fvLength to a vairable and use it in the next line too.

GanShun: call it something like tbpStart or something

pkg/visitors/createfv.go
```diff
@@ -0,0 +1,240 @@
...
+	fv.Blocks[0] = uefi.Block{Size: 4096, Count: uint32(size / 4096)}
+	fv.Blocks[1] = uefi.Block{}
+	// Calculate the HeaderLen field
+	fv.HeaderLen = uint16(uefi.FirmwareVolumeFixedHeaderSize + int(unsafe.Sizeof(uefi.Block{}))*len(fv.Blocks))
+
+	fv.DataOffset = uint64(fv.HeaderLen) // unless we add the extended header
+	fv.Length = size
+
+	if name != nil {
+		// TODO: should the extended header offset be computed ?
```
GanShun: Probably, if we ever want to have weird block sizes, not that I've ever seen a system with more than 2 block entries. This is ok for now.

pkg/visitors/createfv.go
```diff
@@ -0,0 +1,240 @@
...
+
+	if name != nil {
+		// Build the ExtHeader
+		extHeader := new(bytes.Buffer)
+		err = binary.Write(extHeader, binary.LittleEndian, fv.FirmwareVolumeExtHeader)
+		if err != nil {
+			return nil, fmt.Errorf("unable to construct binary extended header of new firmware volume: got %v", err)
+		}
+
+		extHeaderFile, err := uefi.CreatePadFile(uint64(uefi.FileHeaderMinLength + fv.ExtHeaderSize))
```
GanShun: Is the extended header usually an actual file in the FV? I did not realize that. Can you point to some documentation where that's laid out? thanks.

pkg/visitors/createfv.go
JulienVdG: oups...

pkg/visitors/createfv.go
JulienVdG: done

pkg/visitors/createfv.go
JulienVdG: updated the TODO

pkg/visitors/createfv.go
JulienVdG: I found that out by the way it's created in tiano

JulienVdG: https://github.com/tianocore/edk2/blob/master/BaseTools/Source/C/GenFv/GenFvInternalLib.c#L563 the `AddPadFile` func takes the extended header as argument and is used in `GenerateFvImage` here:
https://github.com/tianocore/edk2/blob/master/BaseTools/Source/C/GenFv/GenFvInternalLib.c#L2772

JulienVdG: I couldn't find a reference in the uefi PI spec.

pkg/visitors/createfv.go
JulienVdG: I see ... Shall I leave a TODO here ?
I'll put a comment just warning about that

pkg/visitors/createfv.go
JulienVdG: In the flash image there where this 16 0xff's plus 8 bytes with some values before the actual ext-header content ... so I wondered where they came from ... and found the above mentioned code. The content matched with the Pad File header so that was it.

pkg/visitors/createfv.go
GanShun: Yeah leave a TODO here.

pkg/visitors/createfv.go
GanShun: I see. That's terrible design. Why did they do that -_-. Do you mind explaining it in comments intthe code so when we come back in the future we won't forget? Thanks

GanShun: Just this last one then I think we're good to go

PR #288, opened 2019-10-22
visitors/assemble: regenerate IFD from the parsed content

JulienVdG:
Note: The rest of the sector is kept in place

__The second commit fix the fact that BIOS/ME Regions `FRegion` was not pointing to the ones in the IFD as intended.__

### DIFF
--- a/pkg/uefi/flash.go
+++ b/pkg/uefi/flash.go
@@ -242,8 +242,8 @@ func NewFlashImage(buf []byte) (*FlashImage, error) {
 		return nil, err
 	}
 
-	// FlashRegions
-	frs := f.IFD.Region.FlashRegions
+	// FlashRegions is an array, make a slice to keep reference to it's content
+	frs := f.IFD.Region.FlashRegions[:]
 
 	// BIOS region has to be valid
 	if !frs[RegionTypeBIOS].Valid() {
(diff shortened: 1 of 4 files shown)


pkg/uefi/flash.go
```diff
@@ -242,8 +242,8 @@ func NewFlashImage(buf []byte) (*FlashImage, error) {
 		return nil, err
 	}
 
-	// FlashRegions
-	frs := f.IFD.Region.FlashRegions
+	// FlashRegions is an array, make a slice to manipulate it
+	frs := f.IFD.Region.FlashRegions[:]
```
GanShun: Do you still need to manipulate it? I don't see anything that changes the size in the following code

JulienVdG: So maybe the comment is more "make it a lice to avoid duplicating it" ? what do you think ?

JulienVdG:
Thanks for your encouraging comment, this is appreciated !

pkg/uefi/flash.go
JulienVdG: or better: // FlashRegions is an array, make a slice to keep reference to it's content
which is exactly what we want

GanShun: Either is fine. I missed the original intent of this change. Thanks!

Issue #292, opened 2019-11-11
guid2english sometimes unexpectedly fails


rjoleary:
Fails with the following error message:

    Error copying buffer: transform: short destination buffer

To reproduce:

    guid2english < pkg/knownguids/guids.go

JulienVdG:
I quickly looked, basically `pkg/guid2emglish/transformer.g` `Transform()` only works when the buffer can be processed at once: `src` is small enough that `dst` can hold the transformed text (size is 4096 in transform Reader implementation).
If src is not treated at once, two problems occur:
 1. dst will not hold the transformed text as src will be 4096 already
 2. src can stop in the middle of a GUID, so this GUID will not be replaced if we consume this part of the GUID...

JulienVdG:
#294 is merged, closing

PR #294, opened 2019-11-14
guid2english: fix short destination buffer #292

JulienVdG:
Processing the `src` at once is only valid if `src` contains the full text to
transform.
When the stream is longer, one need to process elements incrementally by
consuming only the beginning of the buffer if there is a risk that the end
contains only partial data.

Also processing the full `src` buffer at once will generate bigger buffer at
`dst` (most of the case the template adds stuff) so treating only the first
element will always fit in `dst`.

### DIFF
--- a/pkg/guid2english/transformer.go
+++ b/pkg/guid2english/transformer.go
@@ -21,6 +21,10 @@ var guidRegex = regexp.MustCompile(
 	"[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}",
 )
 
+var partialguidRegex = regexp.MustCompile(
+	"[-a-fA-F0-9]{1,36}$",
+)
+
 // Mapper converts a GUID to a string.
 type Mapper interface {
 	Map(guid.GUID) []byte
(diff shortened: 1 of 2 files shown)


pkg/guid2english/transformer.go
```diff
@@ -80,21 +84,51 @@ func New(m Mapper) *Transformer {
...
+			return d, s, transform.ErrShortDst
 		}
-		return t.mapper.Map(*g)
-	})
-	if len(transformed) > len(dst) {
-		return 0, 0, transform.ErrShortDst
+		copy(dst, transformed)
+		return len(transformed), len(src), nil
+	}
+	loc := guidRegex.FindIndex(src)
```
GanShun: Doesn't this cause you to miss the partial guid if you have another GUID who's index can be found?

JulienVdG: I don't get your point...
FindIndex will find the first match in src from the beginning of the buffer. a partial GUID can only be at the end, and we don't consume the end of the buffer so next call to Transform by the transform.Reader will give us the remains of src appended with new data (unless at the end of file where atEOF will eventually be set which cover the case of the file containing a partial GUID at its end)

rjoleary: This looks correct to me. The return is `return loc[0] + len(mappedGUID), loc[1], transform.ErrShortSrc`, so only the first guid is consumed. A partial guid wold be left in the source buffer.

GanShun: I think I misunderstood how the transform function is supposed to work, I think this is right

rjoleary:
Thanks for fixing my crappy code!

JulienVdG:
@rjoleary you're welcome, it was an opportunity to learn more of go... and to distract me while I'm supposed to write docs :rofl:

PR #299, opened 2020-01-10
Scanning more stuff, but no Bds things in sight.

rminnich:


pkg/visitors/table.go
```diff
@@ -110,11 +121,15 @@ func (v *Table) printFirmware(f uefi.Firmware, node, name, typez interface{}, of
 	v2.offset = dataOffset
 	v2.curOffset = v2.offset
 
-	if s, ok := f.(*uefi.Section); ok && v.Scan {
-		for g := range knownguids.GUIDs {
-			if bytes.Contains(s.Buf(), g[:]) {
-				fmt.Fprintf(v.W, "%s\t\t%s\n", indent(v.indent), g.String())
+	if v.Scan {
+		switch s := f.(type) {
+		case *uefi.Section:
```
GanShun: Skip firmware volume image sections, and you can also specifically scan files if they are RAW files because those don't contain sections.

PR #303, opened 2020-01-29
uefi: Add boundary checks

xaionaro:
Don't panic on invalid input data in `uefi.Parse()`. Bugs are caught using fuzz-testing.

### DIFF
--- a/pkg/uefi/firmwarevolume.go
+++ b/pkg/uefi/firmwarevolume.go
@@ -232,9 +232,18 @@ func NewFirmwareVolume(data []byte, fvOffset uint64, resizable bool) (*FirmwareV
 		return nil, err
 	}
 
+	// Boundary checks (to return an error instead of panicking)
+	if fv.Length > uint64(len(data)) {
+		return nil, fmt.Errorf("invalid FV length (is greater than the data length): %d > %d",
+			fv.Length, len(data))
+	}
+
 	// Parse the extended header and figure out the start of data
 	fv.DataOffset = uint64(fv.HeaderLen)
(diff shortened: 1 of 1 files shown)


xaionaro:
I will check the errors found by `circleci` tomorrow and I will update the PR.

pkg/uefi/firmwarevolume.go
```diff
@@ -251,6 +255,10 @@ func NewFirmwareVolume(data []byte, fvOffset uint64, resizable bool) (*FirmwareV
 	fv.FVOffset = fvOffset
 
 	// copy out the buffer.
+	if fv.Length >= uint64(len(data)) {
```
JulienVdG: I think you need a strict comparison here:
if fv.Length > uint64(len(data)) {

pkg/uefi/firmwarevolume.go
```diff
@@ -235,6 +235,10 @@ func NewFirmwareVolume(data []byte, fvOffset uint64, resizable bool) (*FirmwareV
 	// Parse the extended header and figure out the start of data
 	fv.DataOffset = uint64(fv.HeaderLen)
 	if fv.ExtHeaderOffset != 0 && uint64(fv.ExtHeaderOffset) < fv.Length-FirmwareVolumeExtHeaderMinSize {
+		if int(fv.ExtHeaderOffset) >= len(data) {
```
JulienVdG: You probably don't need that test if you do the other one `if fv.Length > uint64(len(data))` first as the condition above is more restrictive : 
if fv.Length <= len(data)
and fv.ExtHeaderOffset < fv.Length-FirmwareVolumeExtHeaderMinSize
then fv.ExtHeaderOffset must be < len(data)

unless fv.Length < FirmwareVolumeExtHeaderMinSize so you might want to test for that instead as I guess it will be easier to understand the sanity check

Issue #304, opened 2020-01-31
Discussion: extra copying justification


xaionaro:
Hello.

I'm wondering if there's any specific reason of this copyings:
```
	newBuf := data[:fv.Length]
	fv.buf = make([]byte, fv.Length)
	copy(fv.buf, newBuf)
```

I've removed copying in my sandbox and everything still works. Also:
```
BenchmarkFianoUEFIParse-8      26365509 ns/op         730 B/op    10 allocs/op
```
vs
```
BenchmarkFianoUEFIParse-8      33305961 ns/op    33555673 B/op    13 allocs/op
```

So this copyings creates a lot of extra memory consumption for nothing.

Also in my case this copyings creates some obstacles, because I'm trying to calculate offsets by subtracting one pointer from another.

```
[xaionaro@void fiano]$ go test ./...
ok      github.com/linuxboot/fiano/cmds/fmap    0.097s
ok      github.com/linuxboot/fiano/cmds/fspinfo 0.024s
?       github.com/linuxboot/fiano/cmds/glzma   [no test files]
?       github.com/linuxboot/fiano/cmds/guid2english    [no test files]
?       github.com/linuxboot/fiano/cmds/utk     [no test files]
ok      github.com/linuxboot/fiano/integration  4.587s
ok      github.com/linuxboot/fiano/pkg/compression      6.119s
ok      github.com/linuxboot/fiano/pkg/fmap     0.030s
ok      github.com/linuxboot/fiano/pkg/fsp      0.012s
ok      github.com/linuxboot/fiano/pkg/guid     0.041s
ok      github.com/linuxboot/fiano/pkg/guid2english     0.020s
?       github.com/linuxboot/fiano/pkg/knownguids       [no test files]
ok      github.com/linuxboot/fiano/pkg/uefi     0.004s
?       github.com/linuxboot/fiano/pkg/unicode  [no test files]
?       github.com/linuxboot/fiano/pkg/utk      [no test files]
ok      github.com/linuxboot/fiano/pkg/visitors 6.907s
?       github.com/linuxboot/fiano/scripts/checklicenses        [no test files]
?       github.com/linuxboot/fiano/scripts/namecollect  [no test files]
[xaionaro@void fiano]$ go test ./... -race
ok      github.com/linuxboot/fiano/cmds/fmap    4.040s
ok      github.com/linuxboot/fiano/cmds/fspinfo 1.010s
?       github.com/linuxboot/fiano/cmds/glzma   [no test files]
[...]

PR #305, opened 2020-01-31
uefi: Do not copy if not necessary

xaionaro:
Removed extra copies. It reduces memory consumption, improves performance and simplifies the result (it's now flat).

See the justification discussion: https://github.com/linuxboot/fiano/issues/304

### DIFF
--- a/pkg/uefi/biosregion.go
+++ b/pkg/uefi/biosregion.go
@@ -84,9 +84,7 @@ func NewBIOSRegion(buf []byte, r *FlashRegion, _ FlashRegionType) (Region, error
 		RegionType: RegionTypeBIOS}
 	var absOffset uint64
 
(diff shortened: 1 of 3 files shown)


GanShun:
Let's discuss this a bit more on the issue. I ran into corruption last time without the copies, but we can do it better.

xaionaro:
See https://github.com/linuxboot/fiano/issues/304 -- there's more details that are required to be considered, so the patch will be more complicated.

xaionaro:
@GanShun : A service I implemented based on fiano, consumes too much memory (due to this copies). What you think if we will add support of a special read-only mode?

xaionaro:
Closing the PR in favor to: https://github.com/linuxboot/fiano/pull/326

PR #308, opened 2020-03-04
pkg/visitors: repack must set valid flag in created file

JulienVdG:
The created VolumeImageFile must have the State set as valid.
note: This was done by assemble prior to #263

### DIFF
--- a/pkg/visitors/repack.go
+++ b/pkg/visitors/repack.go
@@ -116,6 +116,9 @@ func createVolumeImageFile(cs *uefi.Section) (*uefi.File, error) {
 	f := &uefi.File{}
 
 	f.Header.Type = uefi.FVFileTypeVolumeImage
+	// set state to valid
+	f.Header.State = 0x07 ^ uefi.Attributes.ErasePolarity
+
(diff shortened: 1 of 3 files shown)


pkg/visitors/repack.go
```diff
@@ -116,6 +116,9 @@ func createVolumeImageFile(cs *uefi.Section) (*uefi.File, error) {
 	f := &uefi.File{}
 
 	f.Header.Type = uefi.FVFileTypeVolumeImage
+	// set state to valid
+	f.Header.State = 0x07 ^ uefi.Attributes.ErasePolarity
```
rminnich: is the 7 defined anywhere or this another mysterious UEFI undocumented magic number?

JulienVdG: I guess
https://github.com/tianocore/edk2/blob/e63d54db952634c2a3a6a24b6c2a4c6ff66f22bd/MdePkg/Include/Pi/PiFirmwareFile.h#L94

If we want to fix that 7 it need to be fixed at every location in utk

JulienVdG: well 3 location (one being commented out)

Issue #311, opened 2020-04-06
`uefi.Parse` hangs


xaionaro:
Hello.

I'm performing a test similar to a fuzz test. And it appears if you run:
```sh
mkdir /tmp/hang-test
cd /tmp/hang-test
cat > main.go <<EOF
package main
import (
        "io/ioutil"
        "os"

        "github.com/linuxboot/fiano/pkg/uefi"
)
func main() {
        b, err := ioutil.ReadAll(os.Stdin)
        if err != nil {
                panic(err)
        }
        b[5742357]++
        _, _ = uefi.Parse(b)
}
EOF
wget 'https://github.com/xaionaro/hack-trash/raw/master/uefi-firmware-example/GALAGOPRO3.fd'
go run main.go < GALAGOPRO3.fd
```
then it hangs and eats a lot of memory :)

```
1 @ 0x4f9ae6 0x4f9a11 0x4f9a11 0x4f9837 0x4f71a2 0x786e55 0x788480 0x785b28 0x7928f9 0x796e67 0x43225e 0x45f9b1
#       0x4f9ae5        encoding/binary.sizeof+0x285                                    /home/xaionaro/.gimme/versions/go1.13.linux.amd64/src/encoding/binary/binary.go:381
#       0x4f9a10        encoding/binary.sizeof+0x1b0                                    /home/xaionaro/.gimme/versions/go1.13.linux.amd64/src/encoding/binary/binary.go:391
#       0x4f9a10        encoding/binary.sizeof+0x1b0                                    /home/xaionaro/.gimme/versions/go1.13.linux.amd64/src/encoding/binary/binary.go:391
#       0x4f9836        encoding/binary.dataSize+0x116                                  /home/xaionaro/.gimme/versions/go1.13.linux.amd64/src/encoding/binary/binary.go:377
#       0x4f71a1        encoding/binary.Read+0x401                                      /home/xaionaro/.gimme/versions/go1.13.linux.amd64/src/encoding/binary/binary.go:231
#       0x786e54        github.com/linuxboot/fiano/pkg/uefi.NewFile+0xf4                /home/xaionaro/go/src/github.com/linuxboot/fiano/pkg/uefi/file.go:409
#       0x78847f        github.com/linuxboot/fiano/pkg/uefi.NewFirmwareVolume+0x70f     /home/xaionaro/go/src/github.com/linuxboot/fiano/pkg/uefi/firmwarevolume.go:279
[...]

PR #312, opened 2020-04-06
fix(firmwarevolume): infinite loop

xaionaro:
Added a break for a loop which was hanging on `file.Header.ExtendedSize == 0`.

I haven't check if the error is returned correctly, yet. I checked only that it helps to avoid the infinite loop described in #311

ITS: https://github.com/linuxboot/fiano/issues/311


xaionaro:
> This loop break doesn't harm the regular case but the question is more what do you expect to get from damaged images ?

I write a tool to analyze damaged images, where failure is better than hanging. And I do not expect to get anything from a damaged volume :)

PR #316, opened 2020-04-16
insert: support insert into non-empty pad file

ambrisko:
When the DXE FW volume is full, allow insertion to work in a non-empty
pad file of the last firmware volume.  This make utk compatible with
other vendor's tool.

Signed-off-by: Doug Ambrisko <ambrisko@ambrisko.com>

### DIFF
--- a/pkg/visitors/assemble.go
+++ b/pkg/visitors/assemble.go
@@ -58,11 +58,12 @@ func (v *Assemble) Visit(f uefi.Firmware) error {
 			// No children, buffer should already contain data.
 			return nil
 		}
+
+		// First pass to check to see if we need to find free space in a pad file.
 		// We assume the buffer already contains the header. We repopulate the header from the buffer
 		// Construct the full buffer.
 		// The FV header is the only thing we've read in so far.
-		fBuf := f.Buf()
-		fBufLen := uint64(len(fBuf))
+		fBufLen := uint64(len(f.Buf()))
 		// The reason I check against f.Length and fBuf instead of the min size is that the volume could
 		// have extended headers.
 		if f.Length < fBufLen {
@@ -71,6 +72,86 @@ func (v *Assemble) Visit(f uefi.Firmware) error {
 		}
 
 		fileOffset := f.DataOffset
+
+		var pad *uefi.File
+		var shrink uint64
+		for _, file := range f.Files {
+			fileBuf := file.Buf()
+			fileLen := uint64(len(fileBuf))
+			if fileLen == 0 {
+				log.Fatal("Unexpected empty file buffer (at least a header was expected): %v", file.Header.GUID)
+			}
+			if file.Header.GUID == *uefi.FFGUID && fileLen > 24 {
+				// found non-empty PAD to check for free space.
+				pad = file
+			}
+
+			// Pad to the 8 byte alignments.
+			alignedOffset := uefi.Align8(fileOffset)
+			// Read out the file alignment requirements
+			if alignBase := file.Header.Attributes.GetAlignment(); alignBase != 1 {
+				hl := file.HeaderLen()
+				// We need to align the data, not the header. This is so terrible.
+				fileDataOffset := uefi.Align(alignedOffset+hl, alignBase)
+				// Calculate the starting offset of the file
+				newOffset := fileDataOffset - hl
+				if gap := (newOffset - alignedOffset); gap >= 8 && gap < uefi.FileHeaderMinLength {
+					// We need to re align to the next boundary cause we can't put a pad file in here.
+					// Who thought this was a good idea?
+					fileDataOffset = uefi.Align(fileDataOffset+1, alignBase)
+					newOffset = fileDataOffset - hl
+				}
+				if newOffset != alignedOffset {
+					// Add a pad file starting from alignedOffset to newOffset
+					alignedOffset = newOffset
+				}
+			}
+			fileOffset = alignedOffset + fileLen
+		}
+
+		// Check if we're out of space.
+		if f.Length < fileOffset && !f.Resizable {
+			// do we have a non-empty pad file?
+			if pad != nil {
+				shrink = fileOffset - f.Length
+				trim := shrink
+				if trim > (uint64(len(pad.Buf()) + 24)) {
+					return fmt.Errorf("out of space in non-empty padding. space available: %v bytes, need apace: %v bytes", len(pad.Buf()), trim)
+				}
+				// remove free space after the header from pad file and recalculate
+				tmp_header := pad.Buf()[:24]
+				tmp_body := pad.Buf()[trim:]
+				tmp_state := tmp_header[23]
+				tmp_size := len(tmp_header) + len(tmp_body)
+				tmp_header[20] = uint8((tmp_size) & 0xff)
+				tmp_header[21] = uint8((tmp_size / 0x100) & 0xff)
+				tmp_header[22] = uint8((tmp_size / 0x10000) & 0xff)
+				tmp_header[16] = 0x0
+				tmp_header[17] = 0x0
+				tmp_header[23] = 0x0
+				tmp_header[16] = uint8((uint16(0x100) - uint16(uefi.Checksum8(tmp_header))) & 0xff)
+				tmp_header[23] = tmp_state
+				tmp_header[17] = 0xaa // Fixed
+				pad.SetBuf(append(tmp_header, tmp_body...))
+			} else {
+				fmt.Printf("no non-empry pad file\n")
+			}
+		}
+
+		// Second pass to check to see if there is enough free space in FV or non-empty pad file
+		// We assume the buffer already contains the header. We repopulate the header from the buffer
+		// Construct the full buffer.
+		// The FV header is the only thing we've read in so far.
+		fBuf := f.Buf()
+		fBufLen = uint64(len(fBuf))
+		// The reason I check against f.Length and fBuf instead of the min size is that the volume could
+		// have extended headers.
+		if f.Length < fBufLen {
+			return fmt.Errorf("buffer read in bigger than FV length!, expected %v got %v bytes",
+				f.Length, fBufLen)
+		}
+
+		fileOffset = f.DataOffset
 		if f.DataOffset != fBufLen {
 			// remove all old file data
 			fBuf = fBuf[:f.DataOffset]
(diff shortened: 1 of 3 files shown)


pkg/visitors/assemble.go
```diff
@@ -71,6 +72,86 @@ func (v *Assemble) Visit(f uefi.Firmware) error {
...
+		var pad *uefi.File
+		var shrink uint64
+		for _, file := range f.Files {
+			fileBuf := file.Buf()
+			fileLen := uint64(len(fileBuf))
+			if fileLen == 0 {
+				log.Fatal("Unexpected empty file buffer (at least a header was expected): %v", file.Header.GUID)
+			}
+			if file.Header.GUID == *uefi.FFGUID && fileLen > 24 {
+				// found non-empty PAD to check for free space.
```
GanShun, line 85: how do you know this pad file is non empty? it could be empty.

pkg/visitors/insert.go
```diff
@@ -121,7 +121,21 @@ func (v *Insert) Visit(f uefi.Firmware) error {
...
+						fileBuf := f.Files[i].Buf()
+						fileLen := uint64(len(fileBuf))
+						if f.Files[i].Header.GUID == *uefi.FFGUID && fileLen > 24 {
+							// found non-empty PAD to check for free space.
+							non_empty_pad = i
+						}
+					}
+					if non_empty_pad >= 0 {
+						i := non_empty_pad
+						f.Files = append(f.Files[:i], append([]*uefi.File{v.NewFile}, f.Files[i:]...)...)
```
GanShun, line 135: Why are you inserting more files whenever you detect a non empty padfile? There are many bioses that have non empty pad files for alignment reasons.

pkg/visitors/assemble.go
```diff
@@ -71,6 +72,86 @@ func (v *Assemble) Visit(f uefi.Firmware) error {
...
+				tmp_header[16] = uint8((uint16(0x100) - uint16(uefi.Checksum8(tmp_header))) & 0xff)
+				tmp_header[23] = tmp_state
+				tmp_header[17] = 0xaa // Fixed
+				pad.SetBuf(append(tmp_header, tmp_body...))
+			} else {
+				fmt.Printf("no non-empry pad file\n")
+			}
+		}
+
+		// Second pass to check to see if there is enough free space in FV or non-empty pad file
```
GanShun, line 141: I'm hesitant to try to find space in non empty pad files in the middle of the volume, because those pad files may be placed there to satisfy alignment requirements, and moving them around can cause things to silently break.

pkg/visitors/assemble.go
```diff
@@ -71,6 +72,86 @@ func (v *Assemble) Visit(f uefi.Firmware) error {
...
+				tmp_header[22] = uint8((tmp_size / 0x10000) & 0xff)
+				tmp_header[16] = 0x0
+				tmp_header[17] = 0x0
+				tmp_header[23] = 0x0
+				tmp_header[16] = uint8((uint16(0x100) - uint16(uefi.Checksum8(tmp_header))) & 0xff)
+				tmp_header[23] = tmp_state
+				tmp_header[17] = 0xaa // Fixed
+				pad.SetBuf(append(tmp_header, tmp_body...))
+			} else {
+				fmt.Printf("no non-empry pad file\n")
```
GanShun, line 137: *empty

pkg/visitors/assemble.go
```diff
@@ -71,6 +72,86 @@ func (v *Assemble) Visit(f uefi.Firmware) error {
...
+		if f.Length < fileOffset && !f.Resizable {
+			// do we have a non-empty pad file?
+			if pad != nil {
+				shrink = fileOffset - f.Length
+				trim := shrink
+				if trim > (uint64(len(pad.Buf()) + 24)) {
+					return fmt.Errorf("out of space in non-empty padding. space available: %v bytes, need apace: %v bytes", len(pad.Buf()), trim)
+				}
+				// remove free space after the header from pad file and recalculate
+				tmp_header := pad.Buf()[:24]
```
GanShun, line 122: Are you trying to create a file or section header here? There are data structures in pkg/uefi/file.go and pkg/uefi/section.go that can create that for you, Please try not to write the offsets manually by addressing the bytes.

pkg/visitors/assemble.go
```diff
@@ -71,6 +72,86 @@ func (v *Assemble) Visit(f uefi.Firmware) error {
...
+				log.Fatal("Unexpected empty file buffer (at least a header was expected): %v", file.Header.GUID)
+			}
+			if file.Header.GUID == *uefi.FFGUID && fileLen > 24 {
+				// found non-empty PAD to check for free space.
+				pad = file
+			}
+
+			// Pad to the 8 byte alignments.
+			alignedOffset := uefi.Align8(fileOffset)
+			// Read out the file alignment requirements
```
GanShun, line 91: Please don't copy this code from further below, if you need to do a two pass, lets talk about moving it out into a helper function.

pkg/visitors/assemble.go
```diff
@@ -71,6 +72,86 @@ func (v *Assemble) Visit(f uefi.Firmware) error {
...
+				}
+			}
+			fileOffset = alignedOffset + fileLen
+		}
+
+		// Check if we're out of space.
+		if f.Length < fileOffset && !f.Resizable {
+			// do we have a non-empty pad file?
+			if pad != nil {
+				shrink = fileOffset - f.Length
```
GanShun, line 116: This will break many bioses. Non empty pad files are often used to align other modules.

pkg/visitors/assemble.go
ambrisko: I'll have to check. what defines an non-empty pad file UEFITool said it was and that is the file that I'm trying to modify.  I'm not trying to modify all pad files but the non-empty pad file that is just before the volume top file.  I need to look again at the details of this since in my use model, only the non-empty pad file is adjusted.  For my example, pad-file are ignored and only the non-empty one is shrunk.

pkg/visitors/insert.go
ambrisko: When I run out of space to insert EFI drivers in the the DXE FV for example, then they are inserted into the last FV and claims space from the non-empty pad file to insert them.  In my use model I do:
     if $UTK -xzPath ${XZ} ${SRC} insert_dxe $file save $DST
when that fails do running out of space in that FV then I do:
     if $UTK -xzPath ${XZ} ${SRC} insert_end PeiCore $file save $DST
I iterate through all of the EFI drivers.   After a successful insert then I move $DST to $SRC. I insert a EFI driver one at time so that when the DXE FV gets full then it tries the
last firmware volume identified via PeiCore.  However, there is no reason why multiple EFI drivers couldn't be installed at the same time.  Note that as EFI drivers are added the pad file is shrunk by the size of the EFI driver.  So then the modified BIOS shows the new EFI driver as a DXE driver in UEFITool and the the non-empy pad file smaller.

pkg/visitors/assemble.go
ambrisko: I'm not moving the pad file around per se, but shrink it to fit in an EFI driver at the original starting address of the pad file.  This way the ending point is the same.  It is critical that we don't change the ending offset or in my use model the BIOS won't boot since the entry point will have a different address :-(  In the last FV, it is laid out as:
   PEI module
   PEC core
   PEI module
   ....
   Freefrom
   ....
   PEI module
   ...
   Pad
   Raw
   ...
   Pad (non-empty) 
   SEC core
then after the insert
   PEI module
   PEC core
   PEI module
   ....
   Freefrom
   ....
   PEI module
   ...
   Pad
   Raw
   ...
   DXE driver -- new EFI driver inserted
   Pad (non-empty) size is now original size - inserted DXE driver
   SEC core
So the ending address of the pad file hasn't changed and the starting address of the inserted DXE is the same as the original pad file.  The starting address of the pad file is now at the end of the inserted DXE driver.

In one of my use cases the original pad file was 1096544 and is now 853640 with the inserted file 242902 according UEFITool Full size values.  If the ending address of the pad file changes then the system won't boot.

pkg/visitors/assemble.go
ambrisko: It's a section header.  I'll look at that.  I thought I asked about how to do this a better way before.  I'd rather not do it this way but didn't see a different way.

pkg/visitors/assemble.go
ambrisko: I'm not sure how that would work.  Are you saying just try to do a normal insert and then on failure try the use the pad file space?  My approach was to look first if there is a space problem and then if there is, do the pad file method since while figuring out the space problem, it gives how much the pad file needs to be reduced.  If there is no pad file space then error out.  If there isn't a space problem then do the normal insert.

pkg/visitors/assemble.go
ambrisko: The intent is not changing the ending address of the pad file since EFI driver is added so the alignment will be unchanged.

pkg/visitors/assemble.go
GanShun: I see, I think I understand your problem a bit better. Is it correct that PEI, DXE and even SEC are all in the same FV in your image? Hmm let me try and cook up something over the weekend to see if I can do this differently

ambrisko: DXE could be in the this FV, typically they are in the DXE volume but when the free space is used up in the DXE FV then they can spill over into the PEI FV.  That FV doesn't have free space by itself but appears to wrap the free space inside the pad file.  I put some utk table dumps at:
     http://www.ambrisko.com/doug/table.orig
     http://www.ambrisko.com/doug/table.customized
The customized one had 3 DXE drivers added with GUID of type
     9F3A0016-AE55-4288-829D-D22FD344C34*
The last one was inserted into the pad file area
@@ -1591,7 +1596,9 @@ Image                                                 
    File           FFFFFFFF-FFFF-FFFF-FFFF-FFFFFFFFFFFF  EFI_FV_FILETYPE_FFS_PAD                0x18
    File           17088572-377F-44EF-8F4E-B09FFF46A070  EFI_FV_FILETYPE_RAW                 0x27028
    File           C3E36D09-8294-4B97-A857-D5288FE33E28  EFI_FV_FILETYPE_RAW                    0x52
-   File           FFFFFFFF-FFFF-FFFF-FFFF-FFFFFFFFFFFF  EFI_FV_FILETYPE_FFS_PAD            0x10bb60
+   File           9F3A0016-AE55-4288-829D-D22FD344C344  EFI_FV_FILETYPE_DRIVER              0x3b4d6
+    Sec                                                 EFI_SECTION_COMPRESSION             0x3b4be
+   File           FFFFFFFF-FFFF-FFFF-FFFF-FFFFFFFFFFFF  EFI_FV_FILETYPE_FFS_PAD             0xd0688
    File           1BA0062E-C779-4582-8566-336AE8F78F09  EFI_FV_FILETYPE_SECURITY_CORE        0x2148
     Sec                                                 EFI_SECTION_RAW                         0x4
     Sec                                                 EFI_SECTION_PE32                     0x20e4

rminnich:
Do we still need this? I kinda think we do but what should we do here?

rminnich:
I am closing this, for lack of progress.

Issue #164, opened 2018-09-14
Add `Position() int64` to the Firmware interface


rjoleary:

JulienVdG:
Does this have something to do with the offset I calculated for the `layout-table` visitor ?
If so this would make the visitor easier, but all visitors that modify the firmware will have to update the fields behind that interface (probably the assemble visitor can do it for them)

xaionaro:
@rjoleary :
Out of curiosity: why `int64` (not `uint64`)?

@JulienVdG : 

> If so this would make the visitor easier, but all visitors that modify the firmware will have to update the fields behind that interface (probably the assemble visitor can do it for them)

Am I understand correctly it means: it will be enough to calculate the absolute offsets, store them within `uefi.Firmware` implementations and update only in `visitors.Assemble`?

JulienVdG:
@xaionaro I think so:
 - an element cannot know it's own offset, it needs to get it from it's parent, so it's probably easier to store it
 - the offset can be calculated when parsing, each container knowing the rules to compute the offsets (if you look at the table visitor you will see that it's not trivial).
 - most visitor that modify the firmware do a call to the assemble visitor, so ensuring assemble updates the offsets could be enough

xaionaro:
@JulienVdG :
We think to use your code of `layout-table` to update offsets. It seems to be less invasive than modify all the constructor functions for this one feature. It's nice when some logic is isolated and does not require to change other logic.

Also it may essentially simplify the 3rd step (to update offsets on assemble).

orangecms:
reviving this :zombie: - I've wanted this all along for https://github.com/fiedka/fiedka :sweat_smile:

PR #317, opened 2020-05-15
WIP: Add Position() to Firmware interface

karlodwyer:
[Original Issue here](https://github.com/linuxboot/fiano/issues/164)

@xaionaro asked on the opensource firmware slack for suggestions around implementing this idea. The approach suggested was to publish a PR and work it out form there, so he and I talked a bit about this.

Some things to note:
* We understand it is not good to export a visitor `PositionUpdater` from uefi, but we are open for discussion on how to fix it.
* This change and `Assemble` probably need more testing to give us more assurance that we aren't regressing anything.

### DIFF
--- a/pkg/uefi/flash.go
+++ b/pkg/uefi/flash.go
@@ -38,6 +38,9 @@ type FlashDescriptor struct {
 
 	//Metadata for extraction and recovery
 	ExtractPath string
+
+	// the Absolute Offset from the beginning of the firmware.
+	AbsOffSet uint64
 }
 
 // FindSignature searches for an Intel flash signature.
@@ -82,6 +85,11 @@ func (fd *FlashDescriptor) Apply(v Visitor) error {
 // ApplyChildren calls the visitor on each child node of FlashDescriptor.
 func (fd *FlashDescriptor) ApplyChildren(v Visitor) error {
 	return nil
+}
+
+// Position returns the absolution position of the node in the firmware image.
+func (fd *FlashDescriptor) Position() uint64 {
+	return fd.AbsOffSet
 }
 
 // ParseFlashDescriptor parses the ifd from the buffer
@@ -139,6 +147,9 @@ type FlashImage struct {
 	// Metadata for extraction and recovery
 	ExtractPath string
 	FlashSize   uint64
+
+	// the Absolute Offset from the beginning of the firmware.
+	AbsOffSet uint64
 }
 
 // Buf returns the buffer.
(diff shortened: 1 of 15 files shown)


JulienVdG:
Hello, first I've not looked into the code yet, but regarding your point about `PositionUpdater` above you can place it in the visitor package along with other visitors.
There is already coupling between the two package anyway (assemble or table cannot work without knowing about each uefi type).
A visitor is not required to provide cli (again assemble is a good example, there is no `RegisterCLI` in assemble.go)
hope this helps.

xaionaro:
> Hello, first I've not looked into the code yet, but regarding your point about `PositionUpdater` above you can place it in the visitor package along with other visitors.

The problem is:
We want to get objects with updates offsets from `uefi.Parse`. So this called is required to be called from `uefi.Parse`. And if we put this visitor to `visitors` we get a circular dependency:
`uefi` -> `visitor` -> `uefi`.

This circular dependency could be broken if we will move types (like `BIOSPadding`) to a separate package (like `./pkg/uefi/types`). In this case the new visitor could be implemented in `./internal/visitors` and used from both packages `uefi` and `visitors`. But since it is changes in the skeleton of the project it would be nice to understand if this approach is approved :)

pkg/uefi/flash.go
```diff
@@ -156,6 +167,11 @@ func (f *FlashImage) ApplyChildren(v Visitor) error {
 	return nil
 }
 
+// Position returns the absolution position of the node in the firmware image.
```
GanShun, line 170: *absolute

pkg/uefi/update_positions.go
```diff
@@ -0,0 +1,99 @@
...
+	offset    uint64
+	curOffset uint64
+}
+
+// Run wraps Visit and performs some setup and teardown tasks.
+func (v *PositionUpdater) Run(f Firmware) error {
+	return f.Apply(v)
+}
+
+// Visit applies the Table visitor to any Firmware type.
```
GanShun, line 22: *PositionUpdater

GanShun (changes requested):
This is an interesting approach, however I'd much rather have the visitor in pkg/visitors. 

The way I would avoid the cyclic dependency is to call updatePositions not from uefi.Parse, but from pkg/utk/utk.go, in the Run function right before other visitors are executed. This also ensures that if we are calling utk on an extracted directory, the positions are also updated. Right now I think it'll only work if you call utk on a bios image.

pkg/uefi/section.go
```diff
@@ -258,6 +261,11 @@ func (s *Section) ApplyChildren(v Visitor) error {
 	return nil
 }
 
+// Position returns the absolution position of the node in the firmware image.
```
GanShun, line 264: absolute*

rminnich:
I am closing this, for lack of interest.

orangecms:
I'm interested, see #321

rminnich:
I don't think this one is going to get finished.

PR #327, opened 2021-02-05
knownguids: add tianocore guids

rminnich:
These are GUIDs from the Mr. Chromebox bug-fix version of tiano core, and/or
Tiano itself.

While there may be some duplicates, it has been useful to keep them blocked
off in their own area, duplicates intact.

The names have (tiano) appended so the origin is easier to see.

This is extremely helpful when debugging Tiano bugs -- a well
with no bottom! -- and working out what of the many UEFI failures
could be the one that finally ended it all. One can just take
the coreboot console log, pipe it through guid2english, and
get some idea of just how bad the problem is. Usually, pretty bad.


GanShun (dismissed):

pkg/knownguids/guids.go
```diff
@@ -10,1224 +10,1547 @@ package knownguids
...
-	*guid.MustParse("52FE8196-F9DE-4D07-B22F-51F77A0E7C41"): "XenTimerDxe",
-	*guid.MustParse("40D1F8F5-4DFE-4E0F-9A15-B1DE9DC9F4ED"): "XenVirtMemInfoLib",
-	*guid.MustParse("338695EA-CA84-4FA2-9DA8-5C4BB87905C6"): "XenioFdtDxe",
-	*guid.MustParse("B7F50E91-A759-412C-ADE4-DCD03E7F7C28"): "XhciDxe",
-	*guid.MustParse("65E5746E-9C14-467D-B5B3-932A66D59F79"): "XhciPei",
-	*guid.MustParse("1831A379-2D48-45BD-9744-D4059D93815D"): "dp",
-	*guid.MustParse("0253F9FA-129A-4A8D-B12E-7DC2B6376302"): "dpDynamicCommand",
-	*guid.MustParse("8DC58D0D-67F5-4B97-9DFC-E442BB9A5648"): "tftp",
-	*guid.MustParse("A487A478-51EF-48AA-8794-7BEE2A0562F1"): "tftpDynamicCommand",
+var GUIDs = map[guid.GUID]string{}
```
rjoleary, line 13: Update the comment at the top of this file about `THIS FILE IS GENERATED! DO NOT MODIFY!` and add instructions on how to update this list.

PR #334, opened 2021-07-29
Add intel me structures from CSS

ChriMarMe:
This PR shall be used to move all intel firmware related structures from 9elements/converged-security-suites to fiano.

@xaionaro @rihter007 will help as well :)


ChriMarMe:
Decided to move package by package, PR by PR.

orangecms (approved):
just squash on merge, please, otherwise all good from my side :)

orangecms (dismissed):

GanShun:
Can we add this parsing to be called from NewMERegion in pkg/uefi/meregion.go? 

Side note we should probably move some of these region descriptors out of pkg/uefi into pkg/intel

GanShun (changes requested):
Can we integrate this with pkg/uefi/meregion.go? I just want the parsing to happen along with everything else.

GanShun:
If the intention is to bring everything over first then integrate after, I'm also ok with that, just let me know if that's what you all wanna do.

orangecms:
I'd suggest that so it becomes easier to see the best way to wire things up.
After all, we're building up giant data structures, I think.

GanShun (approved):
Approving so we get everything in before integrating

pkg/intel/me/me.go
```diff
@@ -0,0 +1,280 @@
...
+	Offset         uint32
+	Length         uint32
+	StartTokens    uint32
+	MaxTokens      uint32
+	ScratchSectors uint32
+	Flags          uint32
+}
+
+func (e FlashPartitionTableEntry) String() string {
+
```
rjoleary, line 99: rm this space

pkg/intel/me/me.go
```diff
@@ -0,0 +1,280 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to get the content of file: %v", err)
+	}
+
+	return uefi, nil
+}
+
+// ParseIntelFirmwareBytes parses the Intel firmware image from bytes
+func ParseIntelFirmwareBytes(imageBytes []byte) (me *IntelME, err error) {
+
```
rjoleary, line 149: rm this space

pkg/intel/me/me.go
```diff
@@ -0,0 +1,280 @@
...
+		// Legacy Header
+		if bytes.HasPrefix(imageBytes[i:], append(make([]byte, 16), []byte(`$FPT`)...)) {
+			legacy = true
+			fptoffset = i
+			break
+		}
+	}
+	if fptoffset == -1 {
+		err = fmt.Errorf("no FlashPartitionTable found")
+		return
```
rjoleary, line 169: Just do a `return nil, fmt.Errorf("no FlashPartitionTable found")` here and elsewhere. See https://tour.golang.org/basics/7 "Naked return statements should be used only in short functions, as with the example shown here. They can harm readability in longer functions."

pkg/intel/me/me.go
```diff
@@ -0,0 +1,280 @@
...
+	fmt.Fprintf(&b, " UMASizeOrReserved  : 0x%x\n", h.UMASizeOrReserved)
+	fmt.Fprintf(&b, " FlashLayoutOrFlags : 0x%x\n", h.FlashLayoutOrFlags)
+	fmt.Fprintf(&b, " Fitc Version       : %d.%d.%d.%d\n", h.FitcMajor, h.FitcMinor, h.FitcHotfix, h.FitcBuild)
+
+	return b.String()
+}
+
+// FlashPartitionTableEntry describes information of a flash partition table entry.
+type FlashPartitionTableEntry struct {
+	Name           [4]uint8
```
rjoleary, line 88: You can create a new type to make this easier to deal with:

```
// String wraps around byte array to give us more control over how strings are
// serialized.
type String struct {
	Value [4]uint8
}

func (s *String) String() string {
	return strings.TrimRight(string(s.Value[:]), "\x00")
}
```

For example, see https://github.com/linuxboot/fiano/blob/master/pkg/fmap/fmap.go#L33

pkg/intel/me/me.go
```diff
@@ -0,0 +1,280 @@
...
+	uefi, err := ParseIntelFirmwareBytes(firmware.Buf())
+	if err != nil {
+		return nil, fmt.Errorf("unable to get the content of file: %v", err)
+	}
+
+	return uefi, nil
+}
+
+// ParseIntelFirmwareBytes parses the Intel firmware image from bytes
+func ParseIntelFirmwareBytes(imageBytes []byte) (me *IntelME, err error) {
```
rjoleary, line 148: Can a unit test be added for this function?

ChriMarMe:
When I get access to images this week I'll write tests for this. Getting it nice and clean before merging :)

pkg/intel/me/me.go
ChriMarMe: Yes it could but I rather reimplement the parsing and remove this function now than implementing a test now

ChriMarMe:
> Yes it could but I rather reimplement the parsing and remove this function now than implementing a test now

:arrow_up: 

Had a change of mind here.

ChriMarMe:
I removed the part which searched for the structure by iterating through a hole image. This should not be the job of a parser.

orangecms (approved):

rjoleary (approved):

ChriMarMe:
I removed the part which searched for the structure by iterating through a hole image. This should not be the job of a parser.

PR #336, opened 2021-08-12
CI - Change linter, increase go version to 1.13

ChriMarMe:
Ci/change linter

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,7 @@ workflows:
 jobs:
   clean-code:
     docker:
-      - image: circleci/golang:1.12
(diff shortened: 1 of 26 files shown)


orangecms:
The case changes in the errors seem a bit weird. 🤔 
Can that be configured?

ChriMarMe:
https://github.com/golang/go/wiki/CodeReviewComments#error-strings

orangecms:
> https://github.com/golang/go/wiki/CodeReviewComments#error-strings

Yeah I'm aware of that, though I wouldn't necessarily agree on the design.
We'll see what we end up with, i.e., if we see sentences starting lowercase now.

PR #340, opened 2021-12-07
Check required size during psp bios tables parsing

rihter007:
Address: https://github.com/9elements/converged-security-suite/issues/301

Function parsePSPFirmware tries to use information from the Embedded Firmware Structure to find both PSP and BIOS tables. But in case of memory scanning for a cookie this could not be the case. Will also add checksum checks.

### DIFF
--- a/pkg/amd/manifest/bios_directory_table.go
+++ b/pkg/amd/manifest/bios_directory_table.go
@@ -55,6 +55,8 @@ type BIOSDirectoryTableEntry struct {
 	DestinationAddress uint64
 }
 
+const BIOSDirectoryTableEntrySize = 16
+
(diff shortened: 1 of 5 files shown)


orangecms (approved):
Thank you, looks good to me!

And I see you also adjusted the messages similar to https://github.com/9elements/converged-security-suite/pull/300/files - thanks! :)

paulmenzel:
Can you please document and reference the firmware image you reproduced this with?

orangecms:
> Can you please document and reference the firmware image you reproduced this with?

Will do! For reproducibility, I'll grab downloadable images from a few vendors like MSI, ASRock, etc, add links, version info, and hashes, plus what directories I verified, plus complementary test scripts. Due to redistribution non-rights, they could only be run locally, I think.

PR #341, opened 2021-12-07
Add checksum calculation for AMD's BIOS and PSP Directories

rihter007:
What a surprise that I found no golang implementation for Fletcher CRC32.


orangecms (approved):
This is something I've wanted to do for a year, but never quite understood the "variants"... So I see now that you're using the examples from https://stackoverflow.com/a/46549752 which are supposedly correct.
Let's get this in now, and I'll see that I can verify the checksums with some real-world examples. Thank you!

PR #348, opened 2021-12-21
Add Intel CBnT manifests parser

xaionaro:
Moving from converged security suite the Intel CBnT manifests parser.

# Good

This package provides structures with parser and compiler of CBnT manifests: boot policy manifest and key manifest.

# Bad

IMHO, the most arguable thing in this package is a codegenerator (I can remove it if required). The idea is that intel structures repeats the same patters over and over, so to have a single source of truth I wrote a bit hacky [codegenerator](https://github.com/9elements/converged-security-suite/tree/master/pkg/intel/metadata/manifest/common/manifestcodegen). The main problem with it is:
* Code quality is not the best. And people complained they do not understand how it works.

The positive side about codegen is that you may remove it at any moment and just continue maintaining structures manually :)


go.mod
```diff
@@ -3,8 +3,15 @@ module github.com/linuxboot/fiano
 go 1.16
 
 require (
+	github.com/dustin/go-humanize v1.0.0
+	github.com/fatih/camelcase v1.0.0
+	github.com/fatih/structtag v1.2.0 // indirect
 	github.com/spf13/pflag v1.0.5
+	github.com/stretchr/testify v1.7.0
```
rihter007, line 10: I'm not sure about testify.
Fiano uses "custom" checks in tests

rihter007 (approved):
LGTM
The only thing that worries me is more dependencies, but this could be ok :)

go.mod
xaionaro: Ouh. OK. I can use standard `testing` package instead. Will fix.

orangecms: I'm fine with this. If anyone complains, we can still move on. :)

PR #350, opened 2022-01-06
Add fittool

xaionaro:
Add a tool to manipulate Intel [FIT](https://www.intel.com/content/dam/develop/external/us/en/documents/firmware-interface-table-bios-specification-r1p2p1.pdf) entries.

Short demo:
```
xaionaro@void:~/go/src/github.com/linuxboot/fiano/cmds/fittool$ dd if=/dev/zero of=/tmp/test bs=1M count=1
1+0 records in
1+0 records out
1048576 bytes (1.0 MB, 1.0 MiB) copied, 0.00133497 s, 785 MB/s
xaionaro@void:~/go/src/github.com/linuxboot/fiano/cmds/fittool$ ./fittool show -f /tmp/test
unable to get FIT entries: unable to get FIT table: unable to locate the table coordinates (does the image contain FIT?): invalid the first entry bytes range: 1 error occurred:
	* start index is less than zero: -4293918720

xaionaro@void:~/go/src/github.com/linuxboot/fiano/cmds/fittool$ ./fittool init -f /tmp/test -p 512
xaionaro@void:~/go/src/github.com/linuxboot/fiano/cmds/fittool$ ./fittool show -f /tmp/test
#   | Type                             | Address              | Size     | Version | Checksum valid  | Checksum
---------------------------------------------------------------------------------------------------------------
0   | FITHeaderEntry            (0x00) | 0x2020205f5449465f   | 1        | 0x0100  | true            | 128
xaionaro@void:~/go/src/github.com/linuxboot/fiano/cmds/fittool$ ./fittool add_raw_headers -f /tmp/test --type 2 --address-pointer $((16#100000)) --size $((16#20000)) --is-checksum-valid
xaionaro@void:~/go/src/github.com/linuxboot/fiano/cmds/fittool$ ./fittool show -f /tmp/test
#   | Type                             | Address              | Size     | Version | Checksum valid  | Checksum
---------------------------------------------------------------------------------------------------------------
0   | FITHeaderEntry            (0x00) | 0x2020205f5449465f   | 2        | 0x0100  | true            | 128
1   | SACM                      (0x02) | 0x100000             | 131072   | 0x1000  | true            | 164
xaionaro@void:~/go/src/github.com/linuxboot/fiano/cmds/fittool$ ./fittool add_raw_headers -f /tmp/test
xaionaro@void:~/go/src/github.com/linuxboot/fiano/cmds/fittool$ ./fittool show -f /tmp/test
[...]

PR #354, opened 2022-01-10
fittool rebased

orangecms:
this is #350 rebased


orangecms:
> Misunderstood you. It appears you meant it is required to rebase it. Sorry I would've done it myself if understood you correctly :)

No worries! The one thing I've noticed that is different between the two PRs is the `os.Setenv("GO111MODULE", "off")` hackaround, but it builds just fine without it, so it should be removed eventually. In #350, it was still there; probably from merge conflict resolution. So I'll merge this one and close #350.

PR #359, opened 2022-01-17
Feature/utk/insert pad file

xaionaro:
Adding a feature to insert padding files.
```
xaionaro@void:~/go/src/github.com/linuxboot/fiano$ go run ./cmds/utk/ -erase-polarity=0xFF /tmp/image \
 create-fv $((32 * 1024)) $((32 * 1024-16)) "61C0F511-A691-4F54-974F-B9A42172CE53" \
 insert_pad_end "61C0F511-A691-4F54-974F-B9A42172CE53" $((32 * 1024-256)) \
 save /tmp/new_image
xaionaro@void:~/go/src/github.com/linuxboot/fiano$ pcr0tool printnodes -as-tree /tmp/new_image
________-____-____-____-____________ *uefi.BIOSRegion  0 65536
  ________-____-____-____-____________ *uefi.BIOSPadding  0 0
  61C0F511-A691-4F54-974F-B9A42172CE53 *uefi.FirmwareVolume  32768 32752
    FFFFFFFF-FFFF-FFFF-FFFF-FFFFFFFFFFFF *uefi.File  32888 32512
  ________-____-____-____-____________ *uefi.BIOSPadding  0 0
```

### DIFF
--- a/pkg/visitors/insert.go
+++ b/pkg/visitors/insert.go
@@ -24,32 +24,32 @@ const (
 
 	// These first two specify a firmware volume.
 
-	// InsertFront inserts a file at the beginning of the firmware volume,
+	// InsertTypeFront inserts a file at the beginning of the firmware volume,
 	// which is specified by 1) FVname GUID, or (File GUID/File name) of a file
 	// inside that FV.
-	InsertFront InsertType = iota
-	// InsertEnd inserts a file at the end of the specified firmware volume.
-	InsertEnd
+	InsertTypeFront InsertType = iota
+	// InsertTypeEnd inserts a file at the end of the specified firmware volume.
+	InsertTypeEnd
 
 	// These two specify a File to insert before or after
-	// InsertAfter inserts after the specified file,
+	// InsertTypeAfter inserts after the specified file,
 	// which is specified by a File GUID or File name.
-	InsertAfter
-	// InsertBefore inserts before the specified file.
-	InsertBefore
-	// InsertDXE inserts into the Dxe Firmware Volume. This works by searching
+	InsertTypeAfter
+	// InsertTypeBefore inserts before the specified file.
+	InsertTypeBefore
+	// InsertTypeDXE inserts into the Dxe Firmware Volume. This works by searching
 	// for the DxeCore first to identify the Dxe Firmware Volume.
-	InsertDXE
+	InsertTypeDXE
 
 	// == Not deprecated ==
 
-	// ReplaceFFS replaces the found file with the new FFS. This is used
+	// InsertTypeReplaceFFS replaces the found file with the new FFS. This is used
 	// as a shortcut for remove and insert combined, but also when we want to make
 	// sure that the starting offset of the new file is the same as the old.
-	ReplaceFFS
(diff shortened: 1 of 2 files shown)


pkg/visitors/insert.go
```diff
@@ -186,4 +222,12 @@ func init() {
 		"insert a file before another file", 2, genInsertCLI(InsertBefore))
 	RegisterCLI(insertTypeNames[ReplaceFFS],
 		"replace a file with another file", 2, genInsertCLI(ReplaceFFS))
+	RegisterCLI(insertTypeNames[InsertPadFront],
```
GanShun, line 225: Instead of creating all these types and commands, perhaps we should rethink this and take an extra "positional" argument in the CLI, and then configure the visitor based on that! the position could be an enum that we extend. I understand if you want to just get this in first tho! we can file an issue to consolidate the insert CLI later.

GanShun (approved):

pkg/visitors/insert.go
xaionaro: Good point. Will do.

xaionaro: Fixed.

pkg/visitors/insert.go
```diff
@@ -173,17 +289,82 @@ func genInsertCLI(iType InsertType) func(args []string) (uefi.Visitor, error) {
...
-		"replace a file with another file", 2, genInsertCLI(ReplaceFFS))
+	RegisterCLI(insertTypeNames[InsertTypeInsert],
+		"insert a file", 4, genInsertFileCLI())
+	RegisterCLI(insertTypeNames[InsertTypeReplaceFFS],
+		"replace a file with another file", 2, genInsertRegularFileCLI(InsertTypeReplaceFFS))
+	RegisterCLI(insertTypeNames[InsertTypeFront],
+		"(deprecated) insert a file at the beginning of a firmware volume", 2, genInsertRegularFileCLI(InsertTypeFront))
+	RegisterCLI(insertTypeNames[InsertTypeEnd],
+		"(deprecated) insert a file at the end of a firmware volume", 2, genInsertRegularFileCLI(InsertTypeEnd))
+	RegisterCLI(insertTypeNames[InsertTypeDXE],
```
GanShun, line 364: It's a bit of a shame that InsertDxe is being deprecated, since it was a fast and easy way to add things to the DXE firmware volume without having to specify a search for the DxeCore. I get that it doesn't fit with the 4 argument requirement the others have though. This is fine for now, though we should probably rethink our parser to allow conditional arguments

GanShun (approved):

orangecms:
I'll get this in now. Just one remark from my side: Let's not artifically expand the identifiers. We have context from the file the code is in, no need to add the `*Type*` thing here everywhere. It also makes it hard to use the module as an API, because e.g. serialization to JSON gets these huge things in a pattern somewhat like this:
```json
{
  "foo": {
    "foobar": {
      "foobarbaz": 123
    }
  }
}}
```

whereas I'd rather wish it was
```json
{
  "foo": {
    "bar": {
      "baz": 123
    }
  }
}}
```

I hope that makes sense.

PR #363, opened 2022-02-07
Minor fix for RSAPSS/SHA384 support for signing of CBnT structures

ChriMarMe:
Signed-off-by: Christopher Meis <christopher.meis@9elements.com>

### DIFF
--- a/pkg/intel/metadata/manifest/signature_types.go
+++ b/pkg/intel/metadata/manifest/signature_types.go
@@ -151,7 +151,7 @@ type SignatureDataInterface interface {
 
 	// Verify returns nil if signedData was indeed signed by key pk, and
 	// returns an appropriate error otherwise.
-	Verify(pk crypto.PublicKey, signedData []byte) error
+	Verify(pk crypto.PublicKey, hashAlgo Algorithm, signedData []byte) error
 }
 
 // SignatureRSAPSS is RSAPSS signature bytes.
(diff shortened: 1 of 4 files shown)


pkg/intel/metadata/manifest/signature_types.go
```diff
@@ -46,6 +46,7 @@ func NewSignatureData(
 	}
 	switch signAlgo {
 	case AlgRSAPSS:
+		fmt.Println("AlgRSAPSS invoked")
```
xaionaro, line 49: Garbage

pkg/intel/metadata/manifest/signature_types.go
```diff
@@ -58,6 +59,7 @@ func NewSignatureData(
 			SaltLength: rsa.PSSSaltLengthAuto,
 			Hash:       crypto.SHA384,
 		}
+		fmt.Printf("Key: %v\nData: %v\nhashData: %v\n", rsaPrivateKey, signedData, bpmHash)
```
xaionaro, line 62: Garbage

pkg/intel/metadata/manifest/signature_types.go
ChriMarMe: Yes I know, thats why the the commit is called: [Print stuff to understand what is going wrong](https://github.com/linuxboot/fiano/pull/363/commits/d74ebcebd92a2aac983ce0dae0728f78cf182def)

pkg/intel/metadata/manifest/signature_types.go
ChriMarMe: Same as the other comment.

pkg/intel/metadata/manifest/signature_types.go
```diff
@@ -194,19 +206,32 @@ func (s SignatureRSAASA) String() string {
...
+	switch hashAlgo {
+	case AlgSHA256:
+		hashfunc = crypto.SHA256
+	case AlgSHA384:
+		hashfunc = crypto.SHA384
+	default:
+		return fmt.Errorf("signature verification for RSAASA only supports SHA256 and SHA384")
+	}
+
+	err = rsa.VerifyPKCS1v15(pk, hashfunc, hash, s)
```
xaionaro, line 234: Couldn't we use the signature size to detect the hash algo if it is not defined (if it is AlgNULL)?

ChriMarMe: Spec says keysize (in bit)/ 8 = signature size. Looks like there is no way to derive hash algo from that information.

xaionaro (approved):

PR #365, opened 2022-02-23
CBFS: marshal offset and more debug info

orangecms:


xaionaro (approved):

pkg/cbfs/fns.go
```diff
@@ -113,6 +115,15 @@ func recString(n string, off uint32, typ string, sz uint32, compress string) str
 	return fmt.Sprintf("%-32s 0x%-8x %-24s 0x%-8x %-4s", n, off, typ, sz, compress)
 }
 
+func cleanString(n string) string {
+	return strings.Map(func(r rune) rune {
+		if r != 0xfffd && (unicode.IsPrint(r) || unicode.IsGraphic(r)) {
```
xaionaro, line 120: Could you comment what is 0xfffd

orangecms: Yea that's a fun one:
https://unicodemap.org/details/0xFFFD/index.html

I saw `6362...00000000ffffff...` print that character - maybe Go just assumes Unicode and does fun things, so I printed each rune, and then saw a bunch of `fffd`. Maybe there is also a flag to treat a string as ASCII only. I'd also need to dig more... or we just accept it. :D

xaionaro: > I'd also need to dig more... or we just accept it. :D

Let's just add a comment into the code for now :)

orangecms: Yea, will do - thank you! :)

PR #367, opened 2022-03-02
pkg/amd/firmware: Wrap PSP and BIOS directories in arrays

orangecms:
This allows for returning multiple directories. Up until now, only the first directories
would be returned, while many firmware images contain multiple directories for various
SoC families and models, e.g., for desktop boards where SoCs can be swapped.

Signed-off-by: Daniel Maslowski <info@orangecms.org>

### DIFF
--- a/pkg/amd/manifest/firmware.go
+++ b/pkg/amd/manifest/firmware.go
@@ -17,20 +17,26 @@ type Firmware interface {
 	OffsetToPhysAddr(offset uint64) uint64
 }
 
-// PSPFirmware contains essential parts of the AMD's PSP firmware internals
-type PSPFirmware struct {
-	EmbeddedFirmware      EmbeddedFirmwareStructure
-	EmbeddedFirmwareRange bytes2.Range
-
-	PSPDirectoryLevel1      *PSPDirectoryTable
-	PSPDirectoryLevel1Range bytes2.Range
-	PSPDirectoryLevel2      *PSPDirectoryTable
-	PSPDirectoryLevel2Range bytes2.Range
-
+type BIOSDir struct {
 	BIOSDirectoryLevel1      *BIOSDirectoryTable
 	BIOSDirectoryLevel1Range bytes2.Range
 	BIOSDirectoryLevel2      *BIOSDirectoryTable
 	BIOSDirectoryLevel2Range bytes2.Range
+}
+
+type PSPDir struct {
+	PSPDirectoryLevel1      *PSPDirectoryTable
+	PSPDirectoryLevel1Range bytes2.Range
+	PSPDirectoryLevel2      *PSPDirectoryTable
+	PSPDirectoryLevel2Range bytes2.Range
+}
+
+// PSPFirmware contains essential parts of the AMD's PSP firmware internals
+type PSPFirmware struct {
+	EmbeddedFirmware      EmbeddedFirmwareStructure
+	EmbeddedFirmwareRange bytes2.Range
+	BIOSDirectories       []BIOSDir
+	PSPDirectories        []PSPDir
 }
 
 // AMDFirmware represents an instance of firmware that exposes AMD specific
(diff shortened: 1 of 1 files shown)


rihter007:
A slice doesn't provide information about the origin of the element.
It is unclear what element should I use, if I end up having multiple.

orangecms:
> A slice doesn't provide information about the origin of the element. It is unclear what element should I use, if I end up having multiple.

Yea, that is also currently the problem: I don't know what I got, I only got a single directory, and I am missing the other directories. From the code I could understand that I got the first directory that is found. And it could be either from the EmbeddedFirmwareStruct or it could have been found by scanning. When the latter is the case, how would we know the processor family/model range?

To expost that information when we have it, I would suggest adding metadata to the respective directory, such as processor family / model range based on the offset. However, that is a different concern; this PR here is to enable multiple directories for a start. WDYT?

GanShun (approved):
I do think this needs to be cleaned up long term, descriptors need to be added to differentiate the PSPDirectories, but it's possible this could be placed in the PSPDir struct instead. I'm ok with merging this right now as it doesn't make anything worse, and looking at the code like this actually warns me that there can be multiple PSP directories.

Perhaps a TODO or block comment is in order warning us that only the first one is handled right now?

rihter007:
> > A slice doesn't provide information about the origin of the element. It is unclear what element should I use, if I end up having multiple.
> 
> Yea, that is also currently the problem: I don't know what I got, I only got a single directory, and I am missing the other directories. From the code I could understand that I got the first directory that is found. And it could be either from the EmbeddedFirmwareStruct or it could have been found by scanning. When the latter is the case, how would we know the processor family/model range?
> 
> To expost that information when we have it, I would suggest adding metadata to the respective directory, such as processor family / model range based on the offset. However, that is a different concern; this PR here is to enable multiple directories for a start. WDYT?

Sorry, for a very long delay. There is a documentation problem, each "directory" has its own purpose. The future code should clearly specify the purpose of each directory. (which we currently can't do)

I agree that the current code assumes, that there could be no more than a single directory in firmware, which could be incorrect.

I propose to leave everything as is, but move "scanning" into a separate function

rihter007 (changes requested):
There is a documentation problem, each "directory" has its own purpose. The future code should clearly specify the purpose of each directory. (which we currently can't do)

I agree that the current code assumes, that there could be no more than a single directory in firmware, which could be incorrect.

I propose to leave everything as is, but move "scanning" into a separate function, this will make the behaviour of what directory do we have clear.

orangecms:
Okay I will create another implementation then. We discussed this in the Fiano sync call, and the proposal is to rework the directory structs, so that they
- contain the range; it's currently side by side, though belongs to the dir
- get a source field; can be "scan", "efs". "xxxdir", depending on whether it resulted from seeking or from following references via EFS or possibly a level 1 directory (called pointers here in the code; "xxxdir" would need be more precise of course such as "fam-17h_model-00h-0fh_biosdir" or something)

pkg/amd/manifest/firmware.go
```diff
@@ -91,15 +100,18 @@ func parsePSPFirmware(firmware Firmware) (*PSPFirmware, error) {
...
+					result.PSPDirectories[0].PSPDirectoryLevel2 = pspDirectoryLevel2
+					result.PSPDirectories[0].PSPDirectoryLevel2Range.Offset = entry.LocationOrValue
+					result.PSPDirectories[0].PSPDirectoryLevel2Range.Length = length
 				}
 			}
 			break
 		}
 	}
 
+	result.BIOSDirectories = []BIOSDir{}
```
rminnich, line 112: not sure I understand this. you can append to a nil slice. But 112 or 113 seem redundant. But you wanted two elements or ...?

orangecms: I don't know enough Go, will see about your suggestion.

orangecms:
I've been working on this again recently, but in a Rust implementation, based on System76/Jeremey Soller's romulan:
https://github.com/orangecms/romulan/tree/CLI-to-teh-limitz

The current objective is diffing two given images, allowing me to somewhat understand them more.
Up until now, we still don't have much documentation. All I can refer to is other people's reversing, what's in Fiano as of now, and what's in the coreboot docs: https://doc.coreboot.org/soc/amd/psp_integration.html#psp-directory-table-entries

PR #372, opened 2022-03-21
cbfs: Add BootSplashRecord

orangecms:
Signed-off-by: Daniel Maslowski <info@orangecms.org>

### DIFF
--- a/pkg/cbfs/types.go
+++ b/pkg/cbfs/types.go
@@ -247,6 +247,10 @@ type BootBlockRecord struct {
 	File
 }
 
+type BootSplashRecord struct {
+	File
+}
+
 type SPDRecord struct {
 	File
 }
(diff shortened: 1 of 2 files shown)


pkg/cbfs/types.go
```diff
@@ -247,6 +247,10 @@ type BootBlockRecord struct {
 	File
 }
 
+type BootSplashRecord struct {
+	File
```
rminnich, line 251: since you have a GetFile, you may want to make this
file
not
File

This may make the race detector less likely to whine at you?

orangecms: I don't understand... isn't this just the same thing as all the others?

orangecms: and I'd want the File to be simply serializable, at least to JSON

GanShun: Yeah File looks correct especially for JSON

GanShun (approved):

pkg/cbfs/types.go
orangecms: Anything else from you @rminnich? Maybe something on the race detector? I'm not seeing where it may find an issue.

PR #380, opened 2022-05-08
Allow ExtendedSize to be shown in utk JSON

AkechiShiro:
Pull request, allowing `ExtendedSize` for UEFI to be shown in the JSON output.

This helps to fix this issue fiedka/fiedka#76 but could be useful for others tools as well.

### DIFF
--- a/integration/utk_test.go
+++ b/integration/utk_test.go
@@ -6,6 +6,7 @@ package utk_test
 
 import (
 	"bytes"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -66,9 +67,11 @@ func TestExtractAssembleExtract(t *testing.T) {
 
 			// Test paths
 			var (
-				dir1   = filepath.Join(tmpDirT, "dir1")
-				tmpRom = filepath.Join(tmpDirT, "tmp.rom")
-				dir2   = filepath.Join(tmpDirT, "dir2")
+				dir1         = filepath.Join(tmpDirT, "dir1")
+				tmpRom       = filepath.Join(tmpDirT, "tmp.rom")
+				dir2         = filepath.Join(tmpDirT, "dir2")
+				summary1Json = filepath.Join(dir1, "summary.json")
+				summary2Json = filepath.Join(dir2, "summary.json")
 			)
 
 			// Extract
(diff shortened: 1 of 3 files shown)


orangecms:
Yea, that's it! Could you please rephrase or just remove the comment?
Fiano can be used in other tools as well, so it shouldn't suggest any tight coupling. Maybe include something in the commit message rather, along the lines of "This allows for use with other tools".

Thank you! :-)

orangecms:
Ah, and finally, run `gofmt`, the Go formatting tool; its code style is very opinionated, and that's what causes the checks to error. 😅

AkechiShiro:
Thanks for your input on this and help, should I also fix some regression tests which are currently failing due to my change?

orangecms:
So the failing test is the extract+reassemble test `pkg/visitors/extract_test.go`.

In a first step, I am improving the output. Insead of printing the entire buffer, I print the position and the respective byte, #382

orangecms:
I gave it another try in #383 by just marshalling the plain `Size` instead, WDYT?

AkechiShiro:
If we can get to fix the tests using #383, I'm all down for it, but @GanShun should also give us his thoughts on it, I guess if that changes would be alright in the codebase.

integration/utk_test.go
```diff
@@ -94,7 +97,19 @@ func TestExtractAssembleExtract(t *testing.T) {
 					t.Errorf("no files in directory %q", d)
 				}
 			}
-
+			sedRemove := func(path string) {
+				sedCmd := exec.Command("sed", "-i", "/\"ExtendedSize\": [0-9]*.*/d", path)
```
rminnich, line 101: we can't have dependencies on sed in a test, can you change how this is done.

integration/utk_test.go
```diff
@@ -94,7 +97,19 @@ func TestExtractAssembleExtract(t *testing.T) {
 					t.Errorf("no files in directory %q", d)
 				}
 			}
-
+			sedRemove := func(path string) {
```
rminnich, line 100: you can remove createTempDir if you want and use t.TempDIr()

rminnich (changes requested):

integration/utk_test.go
rminnich: what you can do, instead, is use regexp.ReplaceAll; just read the file in, do the replace, write it out.
os.ReadAll
regexp.ReplaceAll
os.WriteFile
but what I'm wondering: if you don't want ExtendedSize in the test, why did you put it in the JSON? If it's not in the JSON, then the struct member will be zero; is that not good enough?

rminnich (commented):
I got a bit lost with some of what you did in this PR, sorry.

AkechiShiro:
I'm sorry as well I forgot I had opened this PR a while ago

rminnich:
Fixed a different way, see #424 

So if you want to add ExtendedSize you can see how.

orangecms:
Just for explanation of the original background: Fiedka was not showing size information, but the issue was elsewhere.
The `Size` is exposed and has been integrated in Fiedka, see https://github.com/fiedka/fiedka/issues/76

AkechiShiro:
So this PR's goal is to add `ExtendedSize` that would be in the UI as well as in the test case, just not using `sed` and `diff` just like in the fix @rminnich mentioned (so that test use only Go Packages)

I guess first thing would be to rebase this PR and solve any conflicts

rminnich:
I think we may no longer need this PR? Daniel?

PR #383, opened 2022-05-11
pkg/uefi/file: marshal file size to JSON

orangecms:


pkg/uefi/file.go
```diff
@@ -173,13 +174,34 @@ const (
...
+		t[1] = b[1]
+	}
+	if len(b) > 2 {
+		t[2] = b[2]
+	}
+	return nil
+}
+
+func (t *ThreeUint8) MarshalJSON() ([]byte, error) {
+	res := uint32(t[0]) + uint32(t[1])<<8 + uint32(t[2])<<16
```
GanShun, line 194: Can you use Read3Size? There's already a helper function that does this.

pkg/uefi/file.go
```diff
@@ -173,13 +174,34 @@ const (
 	FileStateValid FileState = FileStateHeaderConstruction | FileStateHeaderValid | FileStateDataValid
 )
 
+type ThreeUint8 [3]uint8
+
+func (t *ThreeUint8) UnmarshalJSON(b []byte) error {
+	if len(b) == 0 {
```
GanShun, line 180: replace with 
if copy(t[:], b) == 0 {
    return fmt.Errorf(...)
}
return nil

orangecms: done

pkg/uefi/file.go
orangecms: now that you showed me how, yea, thank you :)

GanShun (approved):

PR #385, opened 2022-08-04
Bump version of u-root

rjoleary:
Signed-off-by: Ryan O'Leary <ryan@ryanoleary.ca>

PR #386, opened 2022-08-04
CircleCI: Switch to modern Docker images and bump to Go 1.17

orangecms:
This will get #385 to pass, hopefully :)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,17 +20,16 @@ workflows:
 jobs:
   clean-code:
     docker:
-      - image: circleci/golang:1.13
-    working_directory: /go/src/github.com/linuxboot/fiano
+      - image: cimg/go:1.17
     steps:
       - checkout
       - run:
           name: Install ineffassign
-          command: go get github.com/gordonklaus/ineffassign
+          command: |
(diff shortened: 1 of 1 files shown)


orangecms:
see https://circleci.com/developer/images/image/cimg/go for the new images, now just called `go` instead of `golang`

rjoleary (approved):

rjoleary:
Do you also know if it's possible to schedule circleci to run nightly on the master branch? It looks like it hasn't run for a few months. I tried to set up something in the circleci UI. Let's see if that works.

orangecms:
> Do you also know if it's possible to schedule circleci to run nightly on the master branch? It looks like it hasn't run for a few months. I tried to set up something in the circleci UI. Let's see if that works.



> Do you also know if it's possible to schedule circleci to run nightly on the master branch? It looks like it hasn't run for a few months. I tried to set up something in the circleci UI. Let's see if that works.

I don't know this ad-hoc, just checked the docs:
https://circleci.com/docs/scheduled-pipelines#get-started-with-scheduled-pipelines-in-circleci

Assuming what you set up in the UI is similar to what the curl thing there would do, we might need to add something to the yaml still. I can check again later - or someone else may be quicker. :-)

PR #408, opened 2023-04-12
Cbfs compression

PatrickRudolph:


pkg/cbfs/file.go
```diff
@@ -0,0 +1,190 @@
...
+			_, err := buf.Seek(int64(generic.Size)-int64(binary.Size(generic)), io.SeekCurrent)
+			if err != nil {
+				return nil, err
+			}
+		}
+	}
+}
+
+// Compression returns the algorithm used to compress FData.
+// If no compression attribute as found or on error it returns 'None'
```
orangecms, line 160: nit: "... _is_ found ..."

pkg/cbfs/file.go
```diff
@@ -0,0 +1,190 @@
...
+func (f *File) FindAttribute(t Tag) ([]byte, error) {
+	buf := bytes.NewReader(f.Attr)
+	generic := FileAttr{}
+
+	for {
+		if err := binary.Read(buf, Endian, &generic); err != nil {
+			return nil, err
+		}
+
+		Debug("FindAttribute: Found attribute with tag %x", generic.Tag)
```
orangecms, line 142: For an empty file record, i.e., all-zero (`0`), as is the case for type `Deleted2`, this loop runs excessively and the debug print yields `FindAttribute: Found attribute with tag 0` tons of times.
We can check the file type first and immediately return an error instead. WDYT?

pkg/cbfs/file.go
```diff
@@ -0,0 +1,190 @@
...
+
+		if Tag(generic.Tag) == t {
+			_, _ = buf.Seek(-int64(binary.Size(generic)), io.SeekCurrent)
+
+			ret := make([]byte, generic.Size)
+			err := binary.Read(buf, Endian, &ret)
+			return ret, err
+		} else {
+			_, err := buf.Seek(int64(generic.Size)-int64(binary.Size(generic)), io.SeekCurrent)
```
orangecms, line 151: This is where the issue arises: When encountering a `0`, `generic.Size` is `0` and `binary.Size(generic)` is `-8`, so this will seek by `-8`. Split the calculation out, and ensure to seek by 1 at least:

```go
seek := int64(generic.Size)-int64(binary.Size(generic))
if seek <= 0 {
  seek = 1
}
_, err := buf.Seek(seek, io.SeekCurrent)
```

pkg/cbfs/file.go
PatrickRudolph: Looks like Tag Unused and Unused2 mark the end of the attribute list. Added special handling for those tags.

pkg/cbfs/file.go
```diff
@@ -0,0 +1,196 @@
...
+)
+
+var CbfsHeaderMagicNotFound = errors.New("CBFS header magic doesn't match")
+
+func (f *File) MarshalJSON() ([]byte, error) {
+	return json.Marshal(mFile{
+		Name:  f.Name,
+		Start: f.RecordStart,
+		Size:  f.FileHeader.Size,
+		Type:  f.FileHeader.Type.String(),
```
orangecms, line 25: Let's add `Compression` here (and for `mFile`) as well, then it's consistent with the string representation. :)

orangecms (approved):
:partying_face:

PR #413, opened 2023-05-09
Add AMD packages for APCB and PSB

xaionaro:
Adding two packages:
* `pkg/amd/apcb` which implements parser of "AMD PSP Customization Block".
* `pkg/amd/psb` which implements functions related to AMD Platform Secure Boot.

This was our internal code, but AMD gave a permission to publish this code. In particular:

> Based on AMD review of Meta source code, AMD determined that Meta code did not contain proprietary AMD code.


zaolin (approved):
Looks good to me

PR #394, opened 2022-12-30
Bootguard 1.0 support

zaolin:

PR #417, opened 2023-09-05
fix(intel/metadata/bg): Do not reuse Hash instance.

ansiwen:
PR #394 reintroduced a bug for the BtG 1.0 code, that has been fixed by commit [743961f](https://github.com/linuxboot/fiano/commit/743961f033d5d7d2d1305f9b7325d1204bd8f24c) for the CBnT code already, and that results in the wrong calculation of the Key Manifest Pubkey Hash in PrintKMPubKey(). This change reapplies the same fix for the BtG 1.0 code.

### DIFF
--- a/pkg/intel/metadata/bg/crypto_routines.go
+++ b/pkg/intel/metadata/bg/crypto_routines.go
@@ -31,11 +31,11 @@ const (
 )
 
 var hashInfo = []struct {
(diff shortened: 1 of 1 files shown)


orangecms (approved):

orangecms:
Awesome, thank you for the fixup!

PR #418, opened 2023-10-16
pkg/compression: Implement system brotli compression

iangoegebuer:
This has been tested to decompress and recompress brotli compressed images. So far a brotli compressed OVMF image fails to boot after recompression. It fails at reading the section headers.

### DIFF
--- a/pkg/compression/systembrotli.go
+++ b/pkg/compression/systembrotli.go
@@ -53,7 +53,7 @@ func (c *SystemBROTLI) Encode(decodedData []byte) ([]byte, error) {
 	}
 
 	// This seems to be the buffer size needed by the UEFI decompressor
-	header := []byte{0x00, 0x00, 0x00, 0x02,0,0,0,0}
+	header := []byte{0x00, 0x00, 0x00, 0x02, 0, 0, 0, 0}
 
 	header = append(buf.Bytes(), header...)
 
--- a/pkg/compression/systembrotli.go
+++ b/pkg/compression/systembrotli.go
@@ -1,0 +1,63 @@
+// Copyright 2023 the LinuxBoot Authors. All rights reserved
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package compression
+
+import (
+	"bytes"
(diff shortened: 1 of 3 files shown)


iangoegebuer:
@orangecms this was the diff I was talking about at OSFC

pkg/compression/systembrotli.go
```diff
@@ -0,0 +1,64 @@
...
+	return "BROTLI"
+}
+
+// Decode decodes a byte slice of BROTLI data.
+func (c *SystemBROTLI) Decode(encodedData []byte) ([]byte, error) {
+	// The start of the brotli section contains an 8 byte header describing
+	// the final uncompressed size. The real data starts at 0x10
+
+	cmd := exec.Command(c.brotliPath, "--stdout", "-d")
+	cmd.Stdin = bytes.NewBuffer(encodedData[0x10:])
```
xaionaro, line 29: It looks like there is a pure Go brotli implementation: https://github.com/andybalholm/brotli

It think it makes sense to use it instead of adding a dependency on an external tool.

orangecms: We could do the same as for `xz`, where we'd use the Go implementation if no system binary is available, with an optional CLI argument to provide the system binary path. WDYT?

iangoegebuer: I mostly followed the systemlzma approach. Though systemlzma uses the go decompressor instead of the system one.

Take a peek at lines 55-57. For a proper re-compression we probably do need the go implementation since the brotli binary does not expose the buffer size needed.

I don't have any experience with https://github.com/andybalholm/brotli but I did look at it while working on this change. I shy'd away since it was converted with c2go rather than written in go. It is probably worth working with in the future though.

orangecms: Yikes, we stopped here - let's just take it as it is now.  Improvements can follow later.

orangecms (approved):

pkg/compression/systembrotli.go
orangecms: including getting it to work for real...

PR #420, opened 2024-09-06
NewSignatureData works with generic crypto signer.

werwurm:
Casting the crypto.Signer into concrete crypto.PrivateKey
implementations unnecessarily limits the use of the library to software
implementations provided by the golang crypto libraries.

This patch makes the detection of the algorithm based on the public key
which is exposed through crypto.Signer.Public() and then uses the
interface function Sign to perform the actual signing.

This allows for hiding the implementation behind the crypto.Signer
interface, e.g., for signing with an HSM backed key.

Caveat: This patch does not add support for generic SM2 implementations.
But this is not a regression.

Signed-of-by: Janis Danisevskis <jdanisevskis@aurora.tech>

### DIFF
--- a/pkg/intel/metadata/cbnt/signature_types.go
+++ b/pkg/intel/metadata/cbnt/signature_types.go
@@ -10,6 +10,7 @@ import (
 	"crypto/rsa"
 	"crypto/sha256"
 	"crypto/sha512"
+	"encoding/asn1"
 	"fmt"
 	"math/big"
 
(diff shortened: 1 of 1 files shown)


werwurm:
The clean-code error does not seam to be related to my PR. There seems to be an error while installing one of the checks?

rminnich (approved):

rminnich (approved):

orangecms:
> The clean-code error does not seam to be related to my PR. There seems to be an error while installing one of the checks?

Yes, #423 fixed part of the CI issues. Thank you!

PR #427, opened 2024-09-16
Fuzz

rminnich:
This is mpictor's old fuzz stuff all cleaned up. 

We can squash these commits, but perhaps,  because fuzzing is a bit new, we should leave them as is. I don't know.

Maybe there's no instructive value to keeping the mess separate. 2020 was a long time ago.


rminnich:
I think it's time to drop the golangci-lint test, it's just not a good thing to have in CI in my view.

rminnich:
I'll probably squash these as well, given the random bits we don't want, like Gopkg.lock

binjip978 (approved):
I'm not sure about status of dvyukov/go-fuzz right now, it's half deprecated/alive.

rminnich:
I'm going to push this in. From my looking at the fixes, they are legitimate, regardless of the state of the go-fuzz project. 

I've asked Dmitry for the status of go-fuzz.

mpictor:
Thanks for taking care of this, I'd completely forgotten about that draft PR!

FWIW it shouldn't be too much trouble to rework the fuzz tests to use go's built-in fuzzing, though translating the corpus would probably be more work than it's worth.

PR #428, opened 2024-09-16
remove golangci-lint

rminnich:
It has useful recommendations, but the output is
not great, errors with no indications as to which
file it found the error; and it ignores the go version specified in go.mod.


orangecms:
We'll also need to remove the integration in the GitHub project settings.
I'll look into that.

orangecms:
Let's check also how it behaves locally and if newer versions of tje GitHub integration are available. It may just be a simpler matter of a version bump.
We might be able to supply the Go version, let's see about that as well.

rminnich:
sounds good, whatever seems best. I'm just getting tired of its not quite helpful recommendations. :-)

orangecms:
See #429 - looking good!

The Go version to use is not read from go.mod _intentionally_.
It is the version that is selected/installed in the system running the command, just as if you ran it locally.

The docs https://github.com/golangci/golangci-lint-action tell how to specify both the linter version and the Go version.

PR #429, opened 2024-09-16
golangci: lint with Go 1.21, 1.22, 1.23 and macOS, Linux, Windows

orangecms:


orangecms:
that looks helpful - it tells exactly where the issue is in which file, WDYT @rminnich ?

rminnich:
as long as it does not block pushing stuff in, this is ok.

rminnich:
But this advice is not really that useful:
```
Check failure on line 19 in pkg/intel/metadata/fit/ent_startup_ac_module_entry_test.go


GitHub Actions
/ lint (1.22, macos-latest)
SA1019: rand.Read has been deprecated since Go 1.20 because it shouldn't be used: For almost all use cases, [crypto/rand.Read] is more appropriate. (staticcheck)
```
it's a test. I doubt using rand.Read in a test is a problem. 

So I'm still not quite convinced.

orangecms:
It's a linter after all, not magical.

Telling about deprecation is beneficial imho.

It's not blocking anyway... better to have it and be aware if issues such as unhandled errors etc..

binjip978:
> But this advice is not really that useful:
> 
> ```
> Check failure on line 19 in pkg/intel/metadata/fit/ent_startup_ac_module_entry_test.go
> 
> 
> GitHub Actions
> / lint (1.22, macos-latest)
> SA1019: rand.Read has been deprecated since Go 1.20 because it shouldn't be used: For almost all use cases, [crypto/rand.Read] is more appropriate. (staticcheck)
> ```
> 
> it's a test. I doubt using rand.Read in a test is a problem.
> 
> So I'm still not quite convinced.

It possible to use u-root linter configuration https://github.com/u-root/u-root/blob/main/.golangci.yml where we don't block on deprecation issues. Also compared to u-root there are no so much warnings :)

orangecms:
Then let me grab the config from u-root to make it non-blocking.

PR #435, opened 2024-09-17
CI: add golangci linter config

binjip978:


binjip978:
more or less u-root configuration but without revive

orangecms (approved):

orangecms:
Awesome, thank you!

PR #436, opened 2024-09-17
add a display command.

rminnich:
The display command uses the plan9ports devdraw device, and the 9fans devdraw Go support, to allow a display of the state of the buffer.

Currently, a user can do something like this:

Set up so devdraw command is in $PATH

PLAN9=/Users/rminnich/Documents/plan9 export PLAN9 PATH=$PATH:$PLAN9/bin export PATH

Now run utk

utk  ~/Downloads/SBIOS_P5043_PG535_SKU_893_02.02.03_rel_prod.rom  \
	display start \
	remove_dxes_except list  \
	display end

The result will be two windows, one showing before, and one after.

The plan is that this can get fancier:

utk  ~/Downloads/SBIOS_P5043_PG535_SKU_893_02.02.03_rel_prod.rom  \
	display start \
	remove Http.* \
	display nohttp \
	remove_dxes_except list  \
	display end

Also, note, although we only set up one pixel for each point, devdraw is nice to you: if you grow the window it grows the display of ROM. Also, the border is necessery for OSX with its rounded edges.


rminnich:
image

rminnich:
And, yeah, note, they all look the same. Still not sure why.

orangecms:
> And, yeah, note, they all look the same. Still not sure why.

Can you dump the intermediate states to files and just `diff`?

orangecms:
image

just tried this myself, with an empty list, expecting all the DXEs to go down the drain - which should result in a big change, but it looks the same

orangecms:
@rminnich here is a fixup for the rendering https://github.com/rminnich/fiano/pull/1

orangecms:
image

it looks like the issue is with the `remove_dxes_except` command - with just a simple `remove`, I _do_ see a difference now!

rminnich:
I don't need this any more, I'm done with x86, with luck, for good.

PR #441, opened 2024-09-21
pkg/visitors: remove unnecessary temp directories

binjip978:


binjip978:
missed these two, in my previous PR

orangecms (approved):

PR #443, opened 2024-09-28
Intel ME FPT fixes

orangecms:


pkg/uefi/meregion.go
```diff
@@ -95,15 +95,20 @@ func (e MEPartitionEntry) Type() string {
 
 // FindMEDescriptor searches for an Intel ME FTP signature
 func FindMEDescriptor(buf []byte) (int, error) {
-	if bytes.Equal(buf[16:16+len(MEFTPSignature)], MEFTPSignature) {
+	if bytes.Equal(buf[16:16+len(MEFPTSignature)], MEFPTSignature) {
 		// 16 + 4 since the descriptor starts after the signature
```
rminnich, line 99: why not just call bytes.Index? 

Also, so many people depend on this, it needs to come with a test at this point (you found a bug! bad! bad!) :-)

pkg/uefi/meregion.go
```diff
@@ -95,15 +95,20 @@ func (e MEPartitionEntry) Type() string {
...
-	if bytes.Equal(buf[:len(MEFTPSignature)], MEFTPSignature) {
+	if bytes.Equal(buf[:len(MEFPTSignature)], MEFPTSignature) {
 		// + 4 since the descriptor starts after the signature
-		return len(MEFTPSignature), nil
+		return len(MEFPTSignature), nil
 	}
-	return -1, fmt.Errorf("ME Flash Partition Table signature %#02x not found: first 20 bytes are:\n%s", MEFTPSignature, hex.Dump(buf[:20]))
+	fptOffset := bytes.Index(buf, MEFPTSignature)
+	if fptOffset > 0 {
+		fmt.Printf("FPT Signature @ 0x%x\n", fptOffset)
```
rminnich, line 108: this Printf in a package is a Nono

orangecms: oops leftover from debugging - will take that out

pkg/uefi/meregion.go
orangecms: I'm just extending the cases here. Since there is no public spec to my knowledge (I haven't found any) it's hard to say what would be correct. All this knowledge is based on RE.

There are cases where the assumption that the signature is either at the start or in the second 16 bytes holds, and me_cleaner does the same:
https://chromium.googlesource.com/chromiumos/third_party/coreboot/+/refs/heads/stabilize-12239.46.B/util/me_cleaner/me_cleaner.py#579

Looking at more and newer images though, other cases show up, so MEAnalyzer tries a bunch of variants as well, with different implications:
https://github.com/platomav/MEAnalyzer/blob/6d898a6d3972ccec53099e05c5e84ec2abe79197/MEA.py#L11604

So yea bug or... case unknown up until now.
Would you just list the cases known so far for the test cases? I'll see what I can do.

orangecms: Test added. Tbh I don't see the real value, since I am now testing `bytes.Index()`... Would rather add some fixtures and use those instead. Then I'd be unsure because of ME binary distribution issues.

orangecms:
Removed the now needless dump of the first 20 bytes on error. The signature could be anywhere in the ME region.

orangecms:
... making the linter happy

orangecms:
noticed that the only reference where information was taken from is now gone, so added the archive snapshot and more references

pkg/uefi/meregion_test.go
```diff
@@ -59,3 +59,35 @@ func TestMEName_UnmarshalText(t *testing.T) {
...
+		{"elsewhere", elsewhere, 132},
+		{"nowhere", empty, -1},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			r, e := FindMEDescriptor(test.blob)
+			if r != test.res {
+				t.Errorf("got position %d want %d (%q)", r, test.res, e)
+			}
+			if test.res == -1 && e == nil {
```
rminnich, line 88: you can put an err error member in the structure, and use errors.Is(err, test.err)
That would allow tests that succeed.

rminnich (approved):

PR #445, opened 2024-09-29
add Purley silicon Ignition firmware image + ME FPT parsing test

orangecms:
Redistributale, taken from https://edk2.groups.io/g/devel/message/50921

We can use this as a fixture for testing pkg/uefi/meregion.go


pkg/uefi/meregion_test.go
```diff
@@ -91,3 +92,25 @@ func TestFindFPTSignature(t *testing.T) {
...
+	t.Run("parse ME FPT", func(t *testing.T) {
+		meRegion, err := os.ReadFile("../../data/PurleySiliconBinPkg/MeFirmware/IgnitionFirmware/MeRegion.bin")
+		if err != nil {
+			t.Errorf("%q", err)
+		}
+		expectedEntries := 10
+
+		fpt, err := NewMEFPT(meRegion)
+		if err != nil {
+			t.Errorf("expected no error reading a valid ME FPT")
```
rminnich, line 106: might as well make this 
t.Fatalf("NewMEFPT: got %v, want nil", err)

pkg/uefi/meregion_test.go
```diff
@@ -91,3 +92,25 @@ func TestFindFPTSignature(t *testing.T) {
 		})
 	}
 }
+
+func TestParseFPT(t *testing.T) {
+	t.Run("parse ME FPT", func(t *testing.T) {
+		meRegion, err := os.ReadFile("../../data/PurleySiliconBinPkg/MeFirmware/IgnitionFirmware/MeRegion.bin")
+		if err != nil {
+			t.Errorf("%q", err)
```
rminnich, line 100: I think t.Fatalf("os.Readfile(%q): got %v, want nil", name, err)

pkg/uefi/meregion_test.go
```diff
@@ -91,3 +92,25 @@ func TestFindFPTSignature(t *testing.T) {
 		})
 	}
 }
+
+func TestParseFPT(t *testing.T) {
+	t.Run("parse ME FPT", func(t *testing.T) {
```
rminnich, line 97: name := "../../data/PurleySiliconBinPkg/MeFirmware/IgnitionFirmware/MeRegion.bin"
meRegion, err := os.ReadFile(name)

pkg/uefi/meregion_test.go
orangecms: IIUC correctly, it may be against conventions, though technically, we may get both and error _and_ a result, right? I'd want to keep the test running to the end, in that case, so that it may uncover another possible error.

pkg/uefi/meregion_test.go
orangecms: I went with a more elaborate message. Yea, fatal makes sense here. Thanks!

pkg/uefi/meregion_test.go
```diff
@@ -91,3 +92,26 @@ func TestFindFPTSignature(t *testing.T) {
...
+		fixtureFile := "../../data/PurleySiliconBinPkg/MeFirmware/IgnitionFirmware/MeRegion.bin"
+		meRegion, err := os.ReadFile(fixtureFile)
+		if err != nil {
+			t.Fatalf("could not read test fixture %q", fixtureFile)
+		}
+		expectedEntries := 10
+
+		fpt, err := NewMEFPT(meRegion)
+		if err != nil {
+			t.Errorf("expected no error reading a valid ME FPT, got %v", err)
```
rminnich, line 107: This should be log.Fatalf, you're not going anywhere when this fails.

rminnich, line 107: The general form of these errors messages, is the got, want format, e..g
reading ME FPT: got %v, want nil

pkg/uefi/meregion_test.go
```diff
@@ -91,3 +92,26 @@ func TestFindFPTSignature(t *testing.T) {
...
+		}
+		entries := len(fpt.Entries)
+		if entries != expectedEntries {
+			t.Errorf("expected %d entries, got %d", expectedEntries, entries)
```
rminnich, line 111: ftp.Entries: got %d, want %d
and so on.

rminnich (commented):
very  minor tweaks and this can go in

pkg/uefi/meregion_test.go
orangecms: That is what this question was about:
https://github.com/linuxboot/fiano/pull/445#discussion_r1794210468

orangecms: Yea I saw that in the earlier suggestions and find it... irritating/lossy.
Why not provide the detailed idea in the test - which is the spec of behavior?
What I'm saying here is: "I expect this function to return either a result or an error, but never both".
Anyway, will change it if that's the way to go.

rminnich (approved):

PR #450, opened 2026-04-25
Add CBnT 2.1 and replace code generation in intel/metadata

micgor32:
CBnT 2.1 (MTL and newer) brings breaking changes:
- `PDRS` and `CNBS` are now mandatory element of `PCD` (see d75ae02).
- SACM v5 moves the header to `0x00050004`, given that the R/W methods for all SACM versions are same, the logic is moved out to the generic implementation (see 1800718, lines 357-377).
- BPM header is > `0x21`, thus `0x22` to `0x25` is recognized as `BootGuardVersion.Version21`.

Replace codegen for manifest/structure methods. This is done by moving out the logic to a dedicated struct (`Common`), that is then embed into the remaining structs. These have to provide their layout descriptor on each call to `Common`. This reduces the amount of the code in general in the `intel/metadata`. Additionally, `bg` and `cbnt` are now merged by letting the per-type constructor to return the correct type implementation. Both concepts are documented in `pkg/intel/metadata/README.md` in 5903d58.

Since the amount of (+- lines) changes is huge, I've tried to split commits into one per type + the unit test for it. There are few types that are just moved from either `bg` or `cbnt`, and contribute to the modified lines, but do not contain any meaningful logic change. I've noted such cases in the commit description. 

Renamed `common/unittest` to `common/integration` to avoid confusion with the purpose of it and `*_test.go` in `cbnt`. The integration test is now extended a bit, and its full scope is document in `pkg/intel/metadata/common/integration/README.md` in c96a8bd.


micgor32:
ping @orangecms, @rminnich

orangecms:
> ping @orangecms, @rminnich

seen it! It's a *bit* of a change... :D
Not sure when I can finf the time to review this thoroughly.

I would also want to port the general logic and structs to
https://github.com/platform-system-interface/intel_fw

MDr164:
I'll take a look as well, will take _a bit_ of time though...

pkg/intel/metadata/cbnt/bootpolicy/manifest.go
```diff
@@ -0,0 +1,1128 @@
...
+	if s.PCDE != nil {
+		if err := s.PCDE.Validate(); err != nil {
+			return fmt.Errorf("error on field 'PCDE': %w", err)
+		}
+	}
+
+	return nil
+}
+
+func (s *ManifestCBnT) Layout() []cbnt.LayoutField {
```
rminnich, line 702: Please put a function comment here, thanks.

pkg/intel/metadata/cbnt/bootpolicy/manifest.go
```diff
@@ -0,0 +1,1128 @@
...
+					return nil
+				}
+				return s.PME
+			},
+			Type: cbnt.ManifestFieldSubStruct,
+		},
+		{
+			ID:    6,
+			Name:  "PMSE: Signature",
+			Size:  func() uint64 { return s.PMSE.TotalSize() },
```
rminnich, line 809: I much prefer this one-line structure to the multiline structures above, it makes it far easier to read.
Can you change the ones above to match this format?

pkg/intel/metadata/cbnt/bootpolicy/manifest.go
```diff
@@ -0,0 +1,1128 @@
...
+	ret, err := s.Common.OffsetOf(s, id)
+	if err != nil {
+		return ret, fmt.Errorf("Manifest: %v", err)
+	}
+
+	return ret, nil
+}
+
+// ReadFrom reads the Manifest from 'r' in format defined in the document #575623.
+// Same note as above: this is an exception from the rule of usijg common approach.
```
rminnich: usijg -> using

pkg/intel/metadata/cbnt/bootpolicy/manifest.go
```diff
@@ -0,0 +1,1128 @@
...
+func (s *ManifestCBnT) ReadFrom(r io.Reader) (returnN int64, returnErr error) {
```
rminnich: do you really need the named return parameters? In the code below in many places you use different return values.

pkg/intel/metadata/cbnt/bootpolicy/manifest.go
```diff
@@ -0,0 +1,1128 @@
...
+			}
+		default:
+			return totalN, fmt.Errorf("there is no field with structure ID '%s' in Manifest", structInfo.ID)
+		}
+		totalN += n
+		previousFieldIndex = fieldIndex
+	}
+}
+
+func (s *ManifestCBnT) RehashRecursive() {
```
rminnich, line 952: I think on all these exported functions, you need a comment.
If RehashRecursive is part of an interface (I forgot to look) it can be as simple as
// xyz implements xyz 
or some such.

pkg/intel/metadata/cbnt/bootpolicy/pm.go
```diff
@@ -0,0 +1,292 @@
...
+			Name:  "Data Size",
+			Size:  func() uint64 { return 2 },
+			Value: func() any { return &s.DataSize },
+			Type:  cbnt.ManifestFieldArrayStatic,
+		},
+		{
+			ID:   3,
+			Name: "Data",
+			Size: func() uint64 {
+				size := uint64(binary.Size(uint16(0)))
```
rminnich: I wonder why it can't just
return int64(binary.Size(uint16(0))) + uint64(binary.Size(s.DataSize))

pkg/intel/metadata/cbnt/bootpolicy/signature.go
```diff
@@ -0,0 +1,144 @@
...
+
+// NewSignature returns a new instance of Signature with
+// all default values set.
+func NewSignature(bgv cbnt.BootGuardVersion) (*Signature, error) {
+	switch bgv {
+	case cbnt.Version10:
+		s := &Signature{StructInfo: cbnt.NewStructInfo(bgv)}
+		copy(s.StructInfo.(*cbnt.StructInfoBG).ID[:], []byte(StructureIDSignature))
+		s.StructInfo.(*cbnt.StructInfoBG).Version = 0x10
+		// Recursively initializing a child structure:
```
rminnich, line 29: Is this comment really necessary? 

Is this from AI?

pkg/intel/metadata/cbnt/bootpolicy/txt.go
```diff
@@ -0,0 +1,330 @@
...
+	PwrMBaseOffset uint32        `default:"0xFE000000" json:"txtPwrMBaseOffset,omitempty"`
+	DigestList     cbnt.HashList `json:"txtDigestList"`
+	Reserved3      [3]byte       `require:"0" json:"txtReserved3,omitempty"`
+
+	SegmentCount uint8 `require:"0" json:"txtSegmentCount,omitempty"`
+}
+
+// NewTXT returns a new instance of TXT with
+// all default values set.
+func NewTXT() *TXT {
```
rminnich, line 43: For all these New functions that set non-zero values: 
what if someone does this: s := &TXT{}
and does not call this function? Will the zero value work or cause panics? 
Should there be a test for the zero value?

rminnich (commented):
WOW, this is an impressive and overwhelming PR! I guess it can not be done in smaller chunks, it has to be one thing?

rminnich (commented):
my inclination is that we need to accept this, although I don't understand it all ... and hope nothing will break.

micgor32:
> WOW, this is an impressive and overwhelming PR! I guess it can not be done in smaller chunks, it has to be one thing?

It could be split into smaller ones yes, the thing is, these would not be functional on their own i.e. after applying it would leave the `metadata` packages broken partially. Though if it makes life easier for reviewing, I'll split it :)

pkg/intel/metadata/cbnt/bootpolicy/signature.go
micgor32: It's leftover from the `manifestcodegen` https://github.com/linuxboot/fiano/blob/main/pkg/intel/metadata/cbnt/cbntbootpolicy/signature_manifestcodegen.go#L39 and https://github.com/linuxboot/fiano/blob/main/pkg/intel/metadata/common/manifestcodegen/cmd/manifestcodegen/template_methods.tpl.go#L62.

Not sure whether it makes sense to have it there, but I assumed there was a reason for it.

pkg/intel/metadata/cbnt/bootpolicy/pm.go
micgor32: Right, my bad. In here I just blindly moved `PM.DataTotalSize` logic into the closure (https://github.com/linuxboot/fiano/blob/main/pkg/intel/metadata/cbnt/cbntbootpolicy/pm_manifestcodegen.go#L199)

pkg/intel/metadata/cbnt/bootpolicy/manifest.go
micgor32: I could do something like 
```go
func() uint64 { if s.PME == nil { return 0 } return s.PME.TotalSize() }
```
as oneliner.
But the check have to stay there for all substructs of BPM that are optional. Otherwise things might fail quite bad :D

pkg/intel/metadata/cbnt/bootpolicy/txt.go
micgor32: Didn't thought about that, will check and add a test for that if needed.

pkg/intel/metadata/cbnt/bootpolicy/manifest.go
micgor32: right, similar case as with size in pm.go, my bad. Though for the error it make sense to keep it as named parameter, since it is used in the closure few lines below.

pkg/intel/metadata/cbnt/bootpolicy/txt.go
micgor32: So I have checked two cases of such usage with `bg-prov` of CSS.
changing this [part](https://github.com/9elements/converged-security-suite/blob/cbnt21-refactor/cmd/core/bg-prov/cmd.go#L508) to:
```go
func (g *generateBPMCmdv2) Run(ctx *context) error {
	var b bootguard.BootGuard
	b.Version = cbnt.Version20
	if g.Config != "" {
		err := b.ReadJSON(g.Config)
		if err != nil {
			return err
		}
	} else {
		cbntbpm := &bootpolicy.ManifestCBnT{}
		bp := &bootpolicy.BPMHCBnT{}

		b.VData.CBNTbpm = cbntbpm
		b.VData.CBNTbpm.BPMHCBnT = *bp
```
works fine (i.e. no panics), though the values that would have been set to non-zero by New<*> func's won't be spec compliant.

micgor32: the linked part should be different line... https://github.com/9elements/converged-security-suite/blob/cbnt21-refactor/cmd/core/bg-prov/cmd.go#L668 sorry :)

pkg/intel/metadata/cbnt/bootpolicy/manifest.go
micgor32: done.

pkg/intel/metadata/cbnt/bootpolicy/manifest.go
micgor32: done

pkg/intel/metadata/cbnt/bootpolicy/pm.go
micgor32: done.

pkg/intel/metadata/cbnt/bootpolicy/manifest.go
micgor32: done.

pkg/intel/metadata/cbnt/bootpolicy/manifest.go
micgor32: done

pkg/intel/metadata/cbnt/bootpolicy/manifest.go
micgor32: `go fmt` converts such oneliner back to the multiline statements. So the `ci/circleci: clean-code` would fail.

rminnich: ah no problem then. leave it as whatever gofmt does. I don't recall that it used to do that.

rminnich:
are you folks at 9e using this today to get work done? It is working well for you?
I don't use it day to day any more, and I feel that we need to make sure you can move forward. Plus, micgor32 I think is now our resident expert on this code, so I am inclined to trust him.
WDYT?

pkg/intel/metadata/cbnt/bootpolicy/signature.go
xaionaro: I don't think the codegen-code used type assertions.

micgor32: @xaionaro oh it didn't, I meant the comment on recursively initializing the child structure.

rminnich (approved):
I am approving it, it's good to have a new maintainer and it seems this is important.

orangecms (approved):
🥳✨🧡

MDr164:
> are you folks at 9e using this today to get work done? It is working well for you?

So far, yes

> I don't use it day to day any more, and I feel that we need to make sure you can move forward. Plus, micgor32 I think is now our resident expert on this code, so I am inclined to trust him. WDYT?

Given that I worked with him quite a bit (and he's sitting one office room over) I'm biased to say yes :)

MDr164 (approved):
Given that a bulk of the changes is code-gen I went through the non generated sections and couldn't spot anything grave

PR #451, opened 2026-05-18
Add replaceraw

JohnAZoidberg:
I wanted to replace raw files like Intel VBT in firmware images. Very similar to replacepe32

### DIFF
--- a/pkg/visitors/replaceraw.go
+++ b/pkg/visitors/replaceraw.go
@@ -1,0 +1,91 @@
+// Copyright 2026 the LinuxBoot Authors. All rights reserved
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package visitors
+
+import (
+	"errors"
+	"os"
+
+	"github.com/linuxboot/fiano/pkg/uefi"
+)
+
+// ReplaceRaw replaces EFI_SECTION_RAW sections with NewRaw for all files matching Predicate.
+type ReplaceRaw struct {
+	// Input
+	Predicate func(f uefi.Firmware) bool
+	NewRaw    []byte
+
+	// Output
+	Matches []uefi.Firmware
+}
+
+// Run wraps Visit and performs some setup and teardown tasks.
+func (v *ReplaceRaw) Run(f uefi.Firmware) error {
+	// Run "find" to generate a list of matches to replace.
+	find := Find{
+		Predicate: v.Predicate,
+	}
+	if err := find.Run(f); err != nil {
+		return err
+	}
+
+	v.Matches = find.Matches
+	if len(find.Matches) == 0 {
+		return errors.New("no matches found for replacement")
+	}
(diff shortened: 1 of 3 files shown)


pkg/visitors/replaceraw_test.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func TestReplaceRawErrors(t *testing.T) {
+	f := parseImage(t)
+
+	var tests = []struct {
+		name   string
+		newRaw []byte
+		match  string
+		err    string
```
rminnich, line 51: since we first wrote fiano, Go error handling picked up some of the nice bits from Rust. 
SO: this err string should be err error

pkg/visitors/replaceraw_test.go
```diff
@@ -0,0 +1,76 @@
...
+			pred, err := FindFilePredicate(test.match)
+			if err != nil {
+				t.Fatal(err)
+			}
+			replace := &ReplaceRaw{
+				Predicate: pred,
+				NewRaw:    test.newRaw,
+			}
+			err = replace.Run(f)
+			if err == nil {
```
rminnich, line 69: This should be ! errors.Is(err, test.err) and you can avoid the else in that case.

pkg/visitors/replaceraw.go
```diff
@@ -0,0 +1,91 @@
...
+	if err := find.Run(f); err != nil {
+		return err
+	}
+
+	v.Matches = find.Matches
+	if len(find.Matches) == 0 {
+		return errors.New("no matches found for replacement")
+	}
+	if len(find.Matches) > 1 {
+		return errors.New("multiple matches found! There can be only one. Use find to list all matches")
```
rminnich, line 39: This errors.New should be
return ErrMultilpleMatches, which you can declare as an exported variable, with a commend, right before this function:
var ErrMultipleMatches = errors.New("multiple matches found! There can be only one. Use find to list all matches")

pkg/visitors/replaceraw.go
```diff
@@ -0,0 +1,91 @@
...
+	find := Find{
+		Predicate: v.Predicate,
+	}
+	if err := find.Run(f); err != nil {
+		return err
+	}
+
+	v.Matches = find.Matches
+	if len(find.Matches) == 0 {
+		return errors.New("no matches found for replacement")
```
rminnich, line 36: You want to avoid errors.New in anything but a variable declaration, and you want to wrap standard errors for tests. And you might was well be as informative as possible. 
return fmt.Errorf("no matches found for %v:%w", theThingYouAreTryingToMatchWhichICan'tSee, os.ErrNotexist)

then in your test, below, you can test for errors.Is(test.err), and test.err can be initialized to os.ErrNotExist, and now you're not string matching any more.
Thanks for this useful contribution!

rminnich (changes requested):
This is great, you just need to change some of the error handling in code and test.

